PR #215, opened 2015-02-04
formatted

kevmoo:
Using `dart_style` format on the source


cbracken:
There are a couple patches out for this already but we have no plans to apply them quite yet.

kevmoo:
@cbracken understood, but the core question: is the result in general better than where we started?

cbracken:
Generally yes, with a couple exceptions for which bugs have been filed. The other reason I don't want to apply this now is that a couple of us have large-ish CLs out that we'd rather not do mega-merges on.

cbracken:
Made a bunch of minor comment/code changes, re-formatted and tweaked the output. Landed now!

Issue #207, opened 2015-02-04
Consider deprecating/removing glob - refer folks to glob package


kevmoo:
https://github.com/dart-lang/glob

Focus on what quiver does best?

cbracken:
Agreed - I think this makes sense.

justinfagnani:
Agreed, though Quiver's glob predates the glob package :)

justinfagnani:
I just looked at package:gob and it imports dart:io, so it's not a replacement for this. If that flaw is fixed in package:glob, then I think we could remove it here.

justinfagnani:
Also, a quick micro-benchmark shows that Quiver's Glob is ~4000x faster than package:glob. I don't fully trust that, of course, but it's something to look into.

yjbanov:
Won't fix. See #232

PR #226, opened 2015-02-11
Remove pattern/glob. Use glob package: pub.dartlang.org/packages/glob.

cbracken:
Addresses #207.


justinfagnani:
README.md needs updating too

justinfagnani:
Actually, this doesn't look good to me. package:glob can't be used in the browser.

cbracken:
Agreed - not merging till the performance and dependency issues in package:glob are addressed.

seaneagan:
@cbracken @justinfagnani I filed an issue for the dart:io dependency [here](https://github.com/dart-lang/glob/issues/1).  It would be nice to have a bug for the performance issues as well.  Would one of you either be able to file that, or share what benchmarks you ran?

Issue #228, opened 2015-02-11
Metronome gets out of sync


jtmcdole:
For a periodic timer, the broadcast stream of time will get out of sync if the listeners perform lots of work. This would have been detected if I had written an extra test to make the listeners use FakeAsync's  elapsed sync to show some work.

PR #229, opened 2015-02-11
Fix drifting times  Issue #228

jtmcdole:


jtmcdole:
Not sure why this didn't link to #228

jtmcdole:
ping

yjbanov:
Just to be paranoid, you made sure the new tests fail without the fix? If yes, LGTM.

lib/src/async/metronome.dart
```diff
@@ -98,6 +98,6 @@ class Metronome extends Stream<DateTime> {
   _tickDate() {
     DateTime now = clock.now();
     _controller.add(now);
```
justinfagnani: Inline the local `now` and add a comment so someone like me doesn't wonder wtf is going on here :)

jtmcdole: Will do.

jtmcdole:
@yjbanov, yep. They blew up exactly how I knew they would.

cbracken:
lgtm

cbracken:
Merged as e47cee3

PR #235, opened 2015-03-03
remove used fields and imports

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/google/quiver-dart/235)

### DIFF
--- a/lib/testing/src/equality/equality.dart
+++ b/lib/testing/src/equality/equality.dart
@@ -94,7 +94,7 @@ class _EqualityGroupMatcher extends Matcher {
     });
 
     // Run the test multiple times to ensure deterministic equals
-    for (var run in range(_repetitions)) {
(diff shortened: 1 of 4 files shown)


lib/testing/src/equality/equality.dart
```diff
@@ -94,7 +94,7 @@ class _EqualityGroupMatcher extends Matcher {
     });
 
     // Run the test multiple times to ensure deterministic equals
-    for (var run in range(_repetitions)) {
```
justinfagnani: :(

justinfagnani: `i` is just as unused as `run` here. Who's complaining about this and why have they never written Python?

kevmoo: Analyzer is complaining. Creating a new iterable doesn't seem like something we should model/encourage folks to do

PR #243, opened 2015-03-27
Rewrite async.iteration to use await/async

justinfagnani:


lib/src/async/iteration.dart
```diff
@@ -49,15 +49,13 @@ Future _doWhileAsync(
...
  */
-Future reduceAsync(Iterable iterable, initialValue, AsyncCombiner combine) =>
-    _reduceAsync(iterable.iterator, initialValue, combine);
-
-Future _reduceAsync(Iterator iterator, currentValue, AsyncCombiner combine) {
-  if (iterator.moveNext()) {
-    return combine(currentValue, iterator.current)
-        .then((result) => _reduceAsync(iterator, result, combine));
+Future reduceAsync(Iterable iterable, initialValue, AsyncCombiner combine) async
+{
```
yjbanov: Move `{` to the previous line

justinfagnani: it's > 80 cols... where's the formatter say I should put it?

yjbanov: Hmm, try it and see what it says.

yjbanov:
LGTM

justinfagnani:
I'm not so sure we should merge right away, I was just seeing where the code could be more clear in 1.9. We need to bump the quiver version and SDK constraint. I'm not sure if there are other things we can do in 1.9 (generators?), so maybe this deserves a 1.9 branch...

zoechi:
Should also increase the lower bound for `sdk` to at least `1.9.0`
```yaml
environment:
  sdk: '>=1.9.0 <2.0.0'
```

PR #244, opened 2015-04-08
merge quiver.streams into quiver.async (fixes #177)

### LINKED ISSUE #177: Clarify what should be in quiver.async vs quiver.streams
quiver.streams was supposed to be a `Stream` equivalent of quiver.iterables, but it has things with no equivalent and/or that don't work on a Stream or collection of Streams (like how quiver.iterables works on ... `Iterables`) like `collect` and `StreamBuffer`. It's possible we want to move those to quiver.async.

yjbanov:


cbracken:
lgtm

justinfagnani:
I really think that `streams` and `iterables` should merge instead, since they should have a correspondence like `Iterable` and `Stream` do.

yjbanov:
@justinfagnani Had that been the case in the core libs, it would make sense for quiver to follow suit, but everything is in `dart:async` over there.

justinfagnani:
@yjbanov I don't follow. `Stream` and `Future` are both in `dart:core`. Both `quiver.iterables` and `quiver.streams` would have been a set of extension methods to those classes had we had extension methods. They should have the same functions in each library, with the same semantics, except for `quiver.streams` being the async versions. It makes much more sense for them to be combined into one Pythin-itertools-like library, but with async support.

justinfagnani:
The document for the relevant `quiver.streams` functions even points out this correspondence:

`enumerate` and `concat` represent `Stream` versions of the same-named quiver.iterables methods.

justinfagnani:
Sorry, I brain-oed on the location of `Stream`. Regardless, I still think that conceptually `iterables` and `streams` are a set of iteration tools that stand on their own. What ever logic is used to merge `streams` into `async` should be used to merge `iterables` into `collection`.

yjbanov:
Here's the final mapping we're going for:

`quiver.async` <-> `dart:async`
`quiver.collection` <-> `dart:collection`
`quiver.iterables` <-> `dart:core/Iterable`

Post 1.0.0 the plan is to move `quiver.async` into its own package, so it will be even more disconnected from `quiver.iterables` and might evolve into something much more feature-rich a la RX.

I agree that both iterables and streams are conceptually some sort of _repeatables_, but there are significant differences too (streams have errors and evolve over time, while iterables are more like snapshots), and @cbracken convinced me that consistency with the SDK has a higher value.

cbracken:
If no further comments, let's merge this.

yjbanov:
Added commits to:

- make tests pass with the new `package:test` (it seems `throwsA` no longer works with futures)
- not skip set tests unnecessarily
- make FakeAsync a little more idiomatic

PTAL

cbracken:
re-lgtm!

PR #263, opened 2015-07-01
Deprecate FutureGroup, points folks to pkg/async

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/google/quiver-dart/263)


yjbanov:
We removed `FakeTimer` in the new https://github.com/QuiverDart/quiver_async. As for deprecating `FutureGroup`, we should either deprecate the whole `package:quiver_async` in favor of `package:async` or keep `FutureGroup` here (could simply reexport the `package:async` version).

cbracken:
Was done in #321

PR #267, opened 2015-09-16
move quiver into "quiver" sub-directory

yjbanov:


justinfagnani:
Won't this prevent using quiver as a git dependency? Or does pub have subdirectory sorry now?

yjbanov:
@justinfagnani good point. How important do you think it is, given that turning a git dependency to a path dependency is one `git clone` away?

justinfagnani:
I think it's pretty important for cases where a fix is on master or another branch and someone would like to try it out. Asking to install via git might be a bit much.

zoechi:
What's the reasoning behind this change anyway?

yjbanov:
@zoechi, this is preparation for the split into multiple packages. The idea is we move quiver one level down and create multiple packages at the top level. Then quiver would reexport stuff from these packages.

justinfagnani:
Why not create new repos? Or, I'm sure many things could move into existing packages like async, or collection.

yjbanov:
We'll do new repos on https://github.com/QuiverDart. Closing this.

PR #268, opened 2015-09-26
Add fromIterable factory method to Multimap to better match Map API.

trinarytree:
I only ran dartfmt on multimap.dart, not multimap_test.dart because the latter would have changed too many unrelated lines.


cbracken:
lgtm. Thanks!

PR #269, opened 2015-09-30
added a shortcut for creating an Optional from a nullable

EmmanuelOga:
In general, Optional is used in places where normally we would just rely on null checking. For this reason, the most common use case is probably creating an object from a nullable, and so I add a shortcut to make this use case more terse.


yjbanov:
We are accepting PRs bringing new functionality in https://github.com/QuiverDart/quiver_optional, the new home for quiver optional.

PR #271, opened 2015-10-07
Add zipWith function.

kseo:


yjbanov:
We're accepting new functionality in https://github.com/QuiverDart/quiver_iterables, the new home for quiver iterables.

kseo:
Do you think that this patch is okay? Then I will resubmit the patch to quiver_iterables.

yjbanov:
It certainly makes sense, but I haven't thought about whether it is worth including as part of Quiver. I was going to bring it up at the next triage meeting to see what others think about it. Several things to consider:

- the existing `zip` function takes an arbitrary number of iterables while your proposed `zipWith` function takes exactly 2, which feels inconsistent, but:
- your function could be significantly faster as it does not need to allocate extra lists
- could we provide specialized `zip2`/`zipWith2`, `zip3`/`zipWith3`, _in addition_ to the more general `zip`/`zipWith`? This would be similar to what we do for [hash codes](https://github.com/QuiverDart/quiver_hashcode/blob/master/lib/hashcode.dart).
- for the general case though `zipWith` can be easily implemented via `zip` like this:

```dart
Iterable zipWith(Iterable<Iterable> iterables, dynamic combine(List)) =>
  zip(iterables).map(combine);
```

PR #273, opened 2015-10-20
change iterables to use sync*

hterkelsen:
cc @cbracken @justinfagnani @yjbanov


yjbanov:
What's the minimum SDK version that supports `sync*`? We constrain it to `>=1.8`. Also, we may stop releasing new versions of quiver and instead release individual packages in https://github.com/QuiverDart. Why don't we move this PR to https://github.com/QuiverDart/quiver_iterables?

hterkelsen:
moved it over there and also bumped the sdk lower bound to 1.9, when async/await was first introduced

PR #279, opened 2016-01-11
Add LruSet implementation based on LruMap

### LINKED ISSUE #278: Add Least-Recently Used Set implementation
Just like `LruMap`, a `LruSet` class would contain only the X most recently added/inspected items.

stevenroose:
This implementation internally just uses a LruMap to handle the LRU algorithm.

A full implementation can be found in #280.

Fixes #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ false.
 
 `listsEqual`, `mapsEqual` and `setsEqual` check collections for equality.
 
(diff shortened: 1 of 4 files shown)


yjbanov:
Do we get to choose between this and #280? :)

Either way, let's move all PRs for new functionality to https://github.com/QuiverDart/quiver_collection, the new home for quiver collections.

PR #280, opened 2016-01-11
Add LruSet implementation

### LINKED ISSUE #278: Add Least-Recently Used Set implementation

stevenroose:
This implementation does not rely on LruMap, as opposed to #279. So it is an alternative to it.

Fixes #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ false.
 
 `listsEqual`, `mapsEqual` and `setsEqual` check collections for equality.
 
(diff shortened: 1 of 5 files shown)


yjbanov:
This is new functionality. Let's move it to https://github.com/QuiverDart/quiver_collection, the new home of quiver collections.

stevenroose:
Moved to here: https://github.com/QuiverDart/quiver_collection/pull/9

PR #281, opened 2016-01-11
Add first and last getter to LinkedLruHashMap

stevenroose:
Even though `LinkedHashMap` does not have these explicit getters, I think they should be there because they are useful.

`LinkedHashSet` does have the getters (I think `LinkedHashMap` does not have them, because it cannot import the `Iterable` interface without the need for a new `Pair` class). Every `Linked*` class should expose first and last. I could make a PR for `LinkedHashMap` too..

I considered the names `firstKey` and `lastKey`. Still not entirely sure about the names, though.

I think I will also add the getters to the LruMap interface.


stevenroose:
I just realized that they `keys` getter is documented to have LRU order. That makes more sense.

PR #282, opened 2016-01-11
Non-periodic Timers are not active while executing callback()

### LINKED ISSUE #277: FakeAsync timers don't follow spec
_FakeTimer.isActive tests if the timer is included in the zone's _timers. Compare this to the DartVM:

```dart
  var timer;
  timer = new Timer(const Duration(seconds: 2), () {
    print(timer.isActive);
  });
  print(timer.isActive);
```

DartVM output is ```true``` ```false```. Execution with FakeAsync lease to ```true``` ```true```, which is troublesome when writing tests that depend on the isActive state.

I think we should swap the following two lines to solve this:
https://github.com/google/quiver-dart/blob/master/lib/testing/src/async/fake_async.dart#L239

jtmcdole:
Fixes #277

Both the DartVM and dartpad.dartlang.org mark a timer as not-active
while the callback is executing. This can be used in production code
to test some condition - but fails when using quiver-dart/async.dart.

### DIFF
--- a/lib/testing/src/async/fake_async.dart
+++ b/lib/testing/src/async/fake_async.dart
@@ -236,8 +236,8 @@ class _FakeAsync implements FakeAsync {
       timer._callback(timer);
       timer._nextCall += timer._duration;
     } else {
-      timer._callback();
       _timers.remove(timer);
+      timer._callback();
     }
   }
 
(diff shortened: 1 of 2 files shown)


jtmcdole:
I ran DartFmt on the test file which caused some lines to change above my actual test.

yjbanov:
LGTM. Let's fix it in https://github.com/QuiverDart/quiver_async too.

jtmcdole:
I must not be up on quiver_async - is everything moving over there?

yjbanov:
@jtmcdole yes, quiver has been split up into multiple packages. Fixes like this one are still welcome here, but anything new should go to one of the packages under https://github.com/QuiverDart.

PR #283, opened 2016-01-11
Add TwoWayGeneratingIterable and optimize LruMap.keys.last

stevenroose:
This new iterable supports both generating from initial to last as from last to initial element.

The class can be used to make `LruMap.keys.last` more efficient so that users can have constant time access to the least-recently used element.


yjbanov:
@stevenroose, thanks for the PR. We are in the process of moving everything to https://github.com/QuiverDart. Could you please move your PR to https://github.com/QuiverDart/quiver_iterables?

Issue #285, opened 2016-03-16
Strong-mode analysis errors prevent DDC compilation


ochafik:
The following list of [strong-mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md) errors prevent code that depend on `package:quiver` from being compiled with [DDC](https://github.com/dart-lang/dev_compiler).

(ran ```dartanalyzer --strong `find lib -name '*.dart'` | sort -u``` with SDK version `1.16.0-dev.0.0`)

```
[error] Base class introduces an invalid override. The type of DelegatingIterable.expand (((E) → Iterable<dynamic>) → Iterable<dynamic>) is not a subtype of Iterable<E>.expand (<T>((E) → Iterable<T>) → Iterable<T>). (./lib/src/collection/delegates/list.dart, line 29, col 34)
[error] Base class introduces an invalid override. The type of DelegatingIterable.expand (((E) → Iterable<dynamic>) → Iterable<dynamic>) is not a subtype of Iterable<E>.expand (<T>((E) → Iterable<T>) → Iterable<T>). (./lib/src/collection/delegates/queue.dart, line 29, col 35)
[error] Base class introduces an invalid override. The type of DelegatingIterable.expand (((E) → Iterable<dynamic>) → Iterable<dynamic>) is not a subtype of Iterable<E>.expand (<T>((E) → Iterable<T>) → Iterable<T>). (./lib/src/collection/delegates/set.dart, line 29, col 33)
[error] Base class introduces an invalid override. The type of DelegatingIterable.fold ((dynamic, (dynamic, E) → dynamic) → dynamic) is not a subtype of Iterable<E>.fold (<T>(T, (T, E) → T) → T). (./lib/src/collection/delegates/list.dart, line 29, col 34)
[error] Base class introduces an invalid override. The type of DelegatingIterable.fold ((dynamic, (dynamic, E) → dynamic) → dynamic) is not a subtype of Iterable<E>.fold (<T>(T, (T, E) → T) → T). (./lib/src/collection/delegates/queue.dart, line 29, col 35)
[error] Base class introduces an invalid override. The type of DelegatingIterable.fold ((dynamic, (dynamic, E) → dynamic) → dynamic) is not a subtype of Iterable<E>.fold (<T>(T, (T, E) → T) → T). (./lib/src/collection/delegates/set.dart, line 29, col 33)
[...]

Issue #287, opened 2016-03-20
io_test fails on Windows


emesx:
Some tests fail on Windows, namely:

- _visitDirectory should handle symlinks_
- _visitDirectory should conditionally recurse sub-directories_
- _visitDirectory shout not infinitely recurse on symlink cycles_

SDK `1.16.0-dev.1.0 windows_x64`

Issue #286, opened 2016-03-18
Make checkNotNull (and friends) generic for strong-mode


ochafik:
Related to #285, but this is to help user code that looks like `doSomething(checkNotNull(someArg))`.

Note that [strong-mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md) comes with generic methods with syntax `/*=T*/ checkNotNull/*<T>*/(/*=T*/ value)`.

ochafik:
cc/ @emesx

emesx:
On it, ETA Monday.

cbracken:
Fixed by #290

PR #290, opened 2016-04-04
Fixes the strong-mode analysis errors.

emesx:
This fixes issues #285 and #286

Ran all tests, all tests pass (except the one reported in #287).
SDK used: 1.16.0-dev.3.0

### DIFF
--- a/lib/pattern.dart
+++ b/lib/pattern.dart
@@ -50,7 +50,7 @@ class _MultiPattern extends Pattern {
       {Iterable<Pattern> this.exclude});
 
   Iterable<Match> allMatches(String str, [int start = 0]) {
-    var _allMatches = [];
+    final _allMatches = <Iterable<Match>>[];
     for (var pattern in include) {
       var matches = pattern.allMatches(str, start);
       if (_hasMatch(matches)) {
--- a/lib/pattern.dart
+++ b/lib/pattern.dart
@@ -50,7 +50,7 @@ class _MultiPattern extends Pattern {
       {Iterable<Pattern> this.exclude});
 
   Iterable<Match> allMatches(String str, [int start = 0]) {
-    final _allMatches = <Iterable<Match>>[];
+    final _allMatches = <Match>[];
     for (var pattern in include) {
       var matches = pattern.allMatches(str, start);
       if (_hasMatch(matches)) {
@@ -61,10 +61,10 @@ class _MultiPattern extends Pattern {
             }
           }
         }
-        _allMatches.add(matches);
+        _allMatches.addAll(matches);
       }
     }
-    return _allMatches.expand((x) => x);
+    return _allMatches;
   }
 
   Match matchAsPrefix(String str, [int start = 0]) {
--- a/lib/src/async/future_stream.dart
+++ b/lib/src/async/future_stream.dart
@@ -47,7 +47,7 @@ class FutureStream<T> extends Stream<T> {
         _controller.close();
       }
       _controller = null;
-    });
+    }) as Future<Stream<T>>;
 
     if (broadcast == true) {
       _controller = new StreamController.broadcast(
--- a/lib/src/async/future_stream.dart
+++ b/lib/src/async/future_stream.dart
@@ -34,12 +34,14 @@ part of quiver.async;
  *     return new FutureStream(futureOfStream);
  */
 class FutureStream<T> extends Stream<T> {
+  static /*=T*/ _identity/*<T>*/(/*=T*/ t) => t;
+
   Future<Stream<T>> _future;
   StreamController<T> _controller;
   StreamSubscription _subscription;
 
   FutureStream(Future<Stream<T>> future, {bool broadcast: false}) {
-    _future = future.catchError((e, stackTrace) {
+    _future = future.then(_identity, onError: (e, stackTrace) {
       // Since [controller] is synchronous, it's likely that emitting an error
       // will cause it to be cancelled before we call close.
       if (_controller != null) {
@@ -47,7 +49,7 @@ class FutureStream<T> extends Stream<T> {
         _controller.close();
       }
       _controller = null;
-    }) as Future<Stream<T>>;
+    });
 
     if (broadcast == true) {
       _controller = new StreamController.broadcast(
--- a/lib/src/async/stream_buffer.dart
+++ b/lib/src/async/stream_buffer.dart
@@ -91,7 +91,7 @@ class StreamBuffer<T> implements StreamConsumer<T> {
 
   List<T> _consume(int size) {
     var follower = 0;
-    final ret = new List<T>(size);
+    var ret = new List<T>(size);
     var leftToRead = size;
     while (leftToRead > 0) {
       var chunk = _chunks.first;
--- a/lib/src/async/stream_buffer.dart
+++ b/lib/src/async/stream_buffer.dart
@@ -46,7 +46,7 @@ class UnderflowError extends Error {
  * Throws [UnderflowError] if [throwOnError] is true. Useful for unexpected
  * [Socket] disconnects.
  */
-class StreamBuffer<T> implements StreamConsumer {
+class StreamBuffer<T> implements StreamConsumer<T> {
   List _chunks = [];
   int _offset = 0;
   int _counter = 0; // sum(_chunks[*].length) - _offset
@@ -91,7 +91,7 @@ class StreamBuffer<T> implements StreamConsumer {
 
   List<T> _consume(int size) {
     var follower = 0;
-    var ret = new List(size);
+    final ret = new List<T>(size);
     var leftToRead = size;
     while (leftToRead > 0) {
       var chunk = _chunks.first;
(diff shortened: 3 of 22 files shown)


lib/src/collection/treeset.dart
```diff
@@ -438,7 +436,7 @@ class AvlTreeSet<V> extends TreeSet<V> {
   }
 
   bool remove(Object item) {
-    AvlNode<V> x = _getNode(item);
+    AvlNode<V> x = _getNode(item as V);
```
ochafik: This will throw if item is not a V, which isn't the intended behaviour.

lib/pattern.dart
```diff
@@ -50,7 +50,7 @@ class _MultiPattern extends Pattern {
       {Iterable<Pattern> this.exclude});
 
   Iterable<Match> allMatches(String str, [int start = 0]) {
-    var _allMatches = [];
+    final _allMatches = <Iterable<Match>>[];
```
yjbanov: Or, make it `<Match>[]`, then below change from `_allMatches.add` to `_allMatches.addAll` and remove `expand`.

lib/src/async/future_stream.dart
```diff
@@ -47,7 +47,7 @@ class FutureStream<T> extends Stream<T> {
         _controller.close();
       }
       _controller = null;
-    });
+    }) as Future<Stream<T>>;
```
yjbanov: Curious, why is `catchError` not a generic method? Seems like it could simply be:

```dart
Future/*<S>*/ catchError/*<S>*/(/*=S*/ onError(dynamic error, [dynamic stackTrace]),
                  {bool test(Object error)}) => null;
```

@leafpetersen @vsmenon

yjbanov: On second thought, this may not need `catchError` at all. This should work just as well:

```dart
_future = future.then((Stream<T> s) => s, onError: (e, stackTrace) {
  // Since [controller] is synchronous, it's likely that emitting an error
  // will cause it to be cancelled before we call close.
  if (_controller != null) {
    _controller.addError(e, stackTrace);
    _controller.close();
  }
  _controller = null;
});
```

Would this pass strong mode?

lib/src/async/stream_buffer.dart
```diff
@@ -91,7 +91,7 @@ class StreamBuffer<T> implements StreamConsumer {
 
   List<T> _consume(int size) {
     var follower = 0;
-    var ret = new List(size);
+    final ret = new List<T>(size);
```
yjbanov: I'm a fan of `final` myself, but to stay consistent with the rest of the code let's continue using `var` unless strong mode requires something else.

lib/src/async/stream_buffer.dart
```diff
@@ -149,7 +149,7 @@ class StreamBuffer<T> implements StreamConsumer {
     Completer streamDone = new Completer();
     _sub = stream.listen((items) {
       _chunks.add(items);
-      _counter += items is List ? items.length : 1;
+      _counter += items is List ? (items as List).length : 1;
```
yjbanov: @vsmenon @leafpetersen Can this be inferred?

lib/src/collection/multimap.dart
```diff
@@ -186,12 +186,12 @@ abstract class _BaseMultimap<K, V, C extends Iterable<V>>
   }
 
   void add(K key, V value) {
-    _map.putIfAbsent(key, _create);
+    _map.putIfAbsent(key, () => _create() as C);
```
yjbanov: This allocates a closure for every add. Let's see if we can avoid it. Can `_create` return `C`?

lib/src/collection/multimap.dart
```diff
@@ -186,12 +186,12 @@ abstract class _BaseMultimap<K, V, C extends Iterable<V>>
...
     _add(_map[key], value);
   }
 
   void addValues(K key, Iterable<V> values) {
-    _map.putIfAbsent(key, _create);
+    _map.putIfAbsent(key, () => _create() as C);
```
yjbanov: Same here

lib/src/collection/treeset.dart
yjbanov: `_getNode` feeds `item` into `comparator`, which is generic and demands a `V`. I would therefore go further and make `item` parameter `V` rather than `Object`.

@cbracken What do you think?

lib/src/collection/treeset.dart
```diff
@@ -627,10 +625,10 @@ class AvlTreeSet<V> extends TreeSet<V> {
    * See [Set.retainAll]
    */
   void retainAll(Iterable<Object> elements) {
-    List<V> chosen = [];
+    List<V> chosen = <V>[];
     for (var target in elements) {
       if (contains(target)) {
-        chosen.add(target);
+        chosen.add(target as V);
```
yjbanov: `contains`, similar to `remove` feeds the item to the comparator, which could get confused if it received a wrong type. What do you think of:

```dart
if (target is V && contains(target)) {
  chosen.add(target as V);
}
```

lib/src/collection/treeset.dart
yjbanov: Actually, nvm, we _are_ implementing `Iterable`. In that case we can probably do this:

```dart
if (item is! V) {
  return false;
}

...
```

lib/src/collection/treeset.dart
```diff
@@ -690,7 +688,7 @@ class AvlTreeSet<V> extends TreeSet<V> {
     AvlNode<V> x = _root;
     int compare = 0;
     while (x != null) {
-      compare = comparator(element, x.object);
+      compare = comparator(element as V, x.object);
```
yjbanov: Similar situation here. It seems everywhere we take `Object` we first should check

```dart
if (element is! V) {
  // return false, or otherwise ignore the element
}
```

lib/src/collection/treeset.dart
```diff
@@ -792,14 +790,14 @@ class AvlTreeSet<V> extends TreeSet<V> {
...
-    // Opitimized for sorted sets
+    // Optimized for sorted sets
     if (other is TreeSet) {
       var i1 = iterator;
       var i2 = other.iterator;
       var hasMore1 = i1.moveNext();
       var hasMore2 = i2.moveNext();
       while (hasMore1 && hasMore2) {
-        var c = comparator(i1.current, i2.current);
+        var c = comparator(i1.current, i2.current as V);
```
yjbanov: Same problem here.

test/testing/equality/equality_test.dart
```diff
@@ -220,8 +220,8 @@ main() {
     test('EqualityGroups', () {
       expect({
         'valid groups one': [
-          named('foo').addPeers(['bar']),
-          named('bar').addPeers(['foo'])
+          named('foo')..addPeers(['bar']),
+          named('bar')..addPeers(['foo'])
```
yjbanov: I'm scared! How did this not fail before? @cbracken

lib/src/collection/treeset.dart
```diff
@@ -148,7 +146,7 @@ abstract class _TreeNode<V> {
     if (node.left != null) {
       return node.left.maximumNode;
     }
-    while (node.parent != null && node.parent._left == node) {
+    while (node.parent != null && node.parent.left == node) {
```
yjbanov: Uh oh! Seems like our tests are not reaching this code. @cbracken

lib/src/async/stream_buffer.dart
leafpetersen: I think not, if I read the code correctly, because of the Dart rules about type promotion.  In regular Dart, items is dynamic, and promotion is not allowed to promote from dynamic.  We allow promotion from dynamic in strong mode, but we still don't allow promotion between non-subtype related types, and I believe (if I read this correctly) that items will be inferred to have type T, and T and List are not subtype related, and so no type promotion will be done.

lib/src/async/future_stream.dart
emesx: Yes it does. Changed.

lib/src/collection/multimap.dart
emesx: Sure, way better.

lib/src/collection/treeset.dart
emesx: Yeah, and the cast is then unnecessary.
(It's a shame that the interface methods take a `Iterable<Object>` though).

leafpetersen: There's active work on making the core libraries strong mode friendly, please file any issues or requests here:  https://github.com/dart-lang/sdk/issues .  area-library is the right tag.

lib/src/collection/treeset.dart
emesx: Added the check to the preconditions

lib/src/collection/treeset.dart
```diff
@@ -436,6 +436,8 @@ class AvlTreeSet<V> extends TreeSet<V> {
   }
 
   bool remove(Object item) {
+    if (item is! V) return false;
+
     AvlNode<V> x = _getNode(item as V);
```
ochafik: Shame that the analyzer doesn't infer that... FYI filed https://github.com/dart-lang/sdk/issues/26195

emesx:
Have just rebased to current master.

PR #292, opened 2016-04-11
enforce strong mode

yjbanov:
Submit _after_ https://github.com/google/quiver-dart/pull/290


kevmoo:
We're already good here, right?

cbracken:
Oh wow -- yes we are. Thanks for closing!

PR #295, opened 2016-04-25
Fix two unused variable warnings.

munificent:
This is a PR for: https://codereview.chromium.org/1914783002/

### DIFF
--- a/lib/testing/src/runtime/checked_mode.dart
+++ b/lib/testing/src/runtime/checked_mode.dart
@@ -33,10 +33,12 @@ bool _checkForCheckedMode() {
   Object sentinal = new Object();
   try {
     var i = 1 as dynamic;
-    String string = i;
+    _takeString(i);
     throw sentinal;
   } catch (e) {
     if (e == sentinal) return false;
   }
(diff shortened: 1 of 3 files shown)


lib/testing/src/runtime/checked_mode.dart
```diff
@@ -33,10 +33,12 @@ bool _checkForCheckedMode() {
   Object sentinal = new Object();
   try {
     var i = 1 as dynamic;
-    String string = i;
+    _takeString(i);
```
yjbanov: Perhaps we should switch to something more obvious, like:

```dart
bool isCheckedMode = false;
assert((isCheckedMode = true));
return checked;
```

yjbanov:
LGTM. No pressure to fix the checked mode thing.

lib/testing/src/runtime/checked_mode.dart
munificent: Ha, yes, that would be simpler. I'm not sure what the original code was about, but I tried to stay as close to it as possible.

PR #296, opened 2016-04-26
Remove an unnecessary cast.

munificent:
Also add an analysis_options file to enable strong mode for everyone
who hacks on quiver.

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,2 +1,0 @@
-analyzer:
-  strong-mode: true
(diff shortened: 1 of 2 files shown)


.analysis_options
```diff
@@ -0,0 +1,2 @@
+analyzer:
```
yjbanov: Let's revert this for now. It's broken on the stable channel. I have a separate PR to add this. Will wait until the stable channel catches up.

munificent: Done.

PR #297, opened 2016-04-26
simplify assertCheckedMode

yjbanov:


lib/testing/src/runtime/checked_mode.dart
```diff
@@ -20,25 +20,11 @@ part of quiver.testing.runtime;
  * Otherwise, throws [StateError].
  */
 void assertCheckedMode() {
-  if (_isCheckedMode == null) _isCheckedMode = _checkForCheckedMode();
+  bool isCheckedMode = false;
+  // Assertions are only enabled in checked mode
```
cbracken: `dart --enable_asserts` enables asserts but not checked mode.

yjbanov: Didn't know that. nvm then.

PR #299, opened 2016-08-12
Implement Optional.isAbsent

thso:


thso:
Could anyone either LGTM or refuse this PR?

jtmcdole:
Could you update the !isPresent test later in the file as well?  Also, add a test for coverage.

cbracken:
Doing a round of cleanups on old issues. 

Took a quick look around at similar functionality in languages/core libs. All the other `Optional` and `Option` implementations I took a look at (Java, Scala, Swift) only support one means of getting at presence/absence. Dart differs from a lot of other languages in its support for isEmpty/isNotEmpty, but I'm not convinced isPresent/isAbsent is the way to go.

An alternative we could have gone with would have been to use `isEmpty`/`isNotEmpty` themselves, but I think that ship has probably sailed given the low benefit for the cost of migration. 

Going to close this one, but if people have strong arguments for supporting two ways to check presence, I'd be interested to hear them.

PR #301, opened 2016-08-23
Return default Map type from asMap

cbracken:
The core Map factory constructor/Map literal creates a LinkedHashMap.
For consistency with the core lib, do the same in Multimap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+#### 0.23.0
+   * `Multimap.asMap` now returns an instance of the default Map type,
+     `LinkedHashMap`.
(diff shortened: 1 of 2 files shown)


yjbanov:
Would be nice if we could test it in a non-flaky fashion, but still LGTM.

cbracken:
Going to hold off on this for the moment. I'd like to sweeten the deal a bit and actually provide the same sort of guarantees across the API of `Multimap` that are provided by `LinkedHashMap`. Essentially, that we preserve insertion order in key/value iteration and that `asMap` actually returns a `LinkedHashMap`.

PR #303, opened 2016-09-13
Add retry function to async library

natebosch:
- Add retry.dart with a single top level method `retry`
- Add retry_test.dart with tests for the the new method
- Sort `part` declarations in async.dart
- Sort subtest calls in async/all_tests.dart

### DIFF
--- a/lib/src/async/retry.dart
+++ b/lib/src/async/retry.dart
@@ -1,0 +1,36 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of quiver.async;
+
+const _defaultInterval = const Duration(milliseconds: 100);
+const _defaultTimeout = const Duration(seconds: 60);
+
+/// Runs an asynchronous [task] repeatedly until it succeeds or [timeout] is
+/// reached.
+///
+/// Waits [interval] to perform the retry after a failed attempt.
+Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),
+    {Duration interval: _defaultInterval,
+    Duration timeout: _defaultTimeout}) async {
+  var end = new DateTime.now().add(timeout);
+  while (true) {
+    try {
+      return await task();
+    } catch (error) {
+      if (new DateTime.now().isAfter(end)) rethrow;
+      await new Future.delayed(interval);
+    }
+  }
+}
--- a/lib/src/async/retry.dart
+++ b/lib/src/async/retry.dart
@@ -39,7 +39,7 @@ Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),
       throw lastCaught;
     } catch (error) {
       if (new DateTime.now().isAfter(end)) rethrow;
-      await new Future.delayed(interval);
+      await new Future<Null>.delayed(interval);
     }
   }
 }
--- a/lib/src/async/retry.dart
+++ b/lib/src/async/retry.dart
@@ -21,13 +21,22 @@ const _defaultTimeout = const Duration(seconds: 60);
 /// reached.
 ///
 /// Waits [interval] to perform the retry after a failed attempt.
+///
+/// If the task fails to complete once within the timeout it will resolve as an
+/// error with a TimeoutException. If the task completes with an error at least
+/// once before the timeout and never succeeds it will resolve as an error with
+/// the last exception from a task run that did not time out.
 Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),
     {Duration interval: _defaultInterval,
     Duration timeout: _defaultTimeout}) async {
   var end = new DateTime.now().add(timeout);
+  dynamic lastCaught;
   while (true) {
     try {
-      return await task();
+      return await task().timeout(end.difference(new DateTime.now()));
+    } on TimeoutException catch (_) {
+      if (lastCaught == null) rethrow;
+      throw lastCaught;
     } catch (error) {
       if (new DateTime.now().isAfter(end)) rethrow;
       await new Future.delayed(interval);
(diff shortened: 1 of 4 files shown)


jtmcdole:
Why not follow the simple pattern:

```dart
await Future.doWhile(() async {
  // do the task.
  // if success || timeout; return false
  // await new Future.delayed(interval);  return true;
});
```

natebosch:
retry() returns the result of the task which is more difficult in the Future.doWhile version.

The task must return a boolean you end up with boilerplate to separately grab the result. Essentially every instance of using that pattern would end up looking like retry() implemented with Future.doWhile()

For the record heres what that looks like:

```dart
Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),                                 
    {Duration interval: _defaultInterval,                                        
    Duration timeout: _defaultTimeout}) async {                                  
  var/*=T*/ result;                                                              
  var end = new DateTime.now().add(timeout);                                     
  await Future.doWhile(() async {                                                
    try {                                                                        
      result = await task();                                                     
      return false;                                                              
    } catch (error) {                                                            
      if (new DateTime.now().isAfter(end)) rethrow;                              
      await new Future.delayed(interval);                                        
      return true;                                                               
    }                                                                            
  });                                                                            
  return result;                                                                 
}
```

This version is noisier than the while(true) version.

jtmcdole:
Sure, I wasn't saying you should have the Future.doWhile() in the retry() function as that's overkill. Thanks for pointing out the return value.

lib/src/async/retry.dart
```diff
@@ -0,0 +1,36 @@
...
+/// reached.
+///
+/// Waits [interval] to perform the retry after a failed attempt.
+Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),
+    {Duration interval: _defaultInterval,
+    Duration timeout: _defaultTimeout}) async {
+  var end = new DateTime.now().add(timeout);
+  while (true) {
+    try {
+      return await task();
```
jtmcdole: Is it important to respect the timeout contract here? e.g.

```dart
await task().timeout(end.difference(new DateTime.now()));
```

Which prevents task() for taking longer than timeout.

natebosch: Hmm, good question - I don't think it is. I can imagine that some callers would want to let task() run longer than the timeout - and it's not a terrible amount of boilerplate for other callers to add their own timeout

```dart
var timeout = const Duration(seconds: 1);
await retry(task, timeout: timeout).timeout(timeout);
```

I'm open to arguments the other way though if we're worried that's confusing behavior.

jtmcdole: I think the (timeout: timeout).timeout(timeout) screams that something is wrong. Also it makes for an accidental infinite loop if the user forgets to pass timeout to the api and just ops for:

```dart
await retry(task).timeout(timeout);
```

I think the argument is then: if you want task to run longer, you should increase timeout.

natebosch: After thinking it through I realized the important bit wasn't letting the task() run longer than the timeout, but that it surfaces something more useful than a TimeoutException when fails at least once.

I've added an inner timeout() call like you suggested, added an extra catch to distinguish between timeouts and real task failures, and expanded the Doc comment to explain the behavior. What do you think?

lib/src/async/retry.dart
```diff
@@ -21,13 +21,22 @@ const _defaultTimeout = const Duration(seconds: 60);
...
+///
+/// If the task fails to complete once within the timeout it will resolve as an
+/// error with a TimeoutException. If the task completes with an error at least
+/// once before the timeout and never succeeds it will resolve as an error with
+/// the last exception from a task run that did not time out.
 Future/*<T>*/ retry/*<T>*/(Future/*<T>*/ task(),
     {Duration interval: _defaultInterval,
     Duration timeout: _defaultTimeout}) async {
   var end = new DateTime.now().add(timeout);
+  dynamic lastCaught;
```
jtmcdole: var lastCaught?

lib/src/async/retry.dart
```diff
@@ -21,13 +21,22 @@ const _defaultTimeout = const Duration(seconds: 60);
...
   while (true) {
     try {
-      return await task();
+      return await task().timeout(end.difference(new DateTime.now()));
+    } on TimeoutException catch (_) {
+      if (lastCaught == null) rethrow;
```
jtmcdole: You are not updating lastCaught, so it'll always be null. rethrow will pull the eject switch, so the next line will never run.  What happens if task() throws a timeout exception - the rethrow will occur too soon, breaking the contract.

test/async/retry_test.dart
```diff
@@ -46,5 +48,12 @@ void main() {
               timeout: const Duration(milliseconds: 1)),
           throwsA(equals('error')));
     });
+
+    test('should timeout long tasks', () async {
+      var task = () => new Completer().future;
+
+      expect(retry(task, timeout: const Duration(milliseconds: 100)),
```
jtmcdole: You should also look at using FakeAsync when writing tests that depend on time, granted your test here is perfectly OK because the "task" is always going to wait forever.

lib/src/async/retry.dart
```diff
@@ -21,13 +21,22 @@ const _defaultTimeout = const Duration(seconds: 60);
 /// reached.
 ///
 /// Waits [interval] to perform the retry after a failed attempt.
+///
+/// If the task fails to complete once within the timeout it will resolve as an
+/// error with a TimeoutException. If the task completes with an error at least
+/// once before the timeout and never succeeds it will resolve as an error with
+/// the last exception from a task run that did not time out.
```
jtmcdole: I'm torn here on this contract. Does it make more sense / remain simple to just throw a timeout exception and not cache the last exception?

natebosch: The use case we've had so far has been in integration tests where getting the actual exception at the end is critical. I think it makes sense to have a distinction a timeout (the task took too long the first time) and other failures.

lib/src/async/retry.dart
natebosch: dynamic is more future-proof if we decide to turn on --no-implicit-dynamic on the analyzer (which I hope we do). To that end I also added a generic argument to the Future.delayed below. I wasn't concerned with --no-implicit-dynamic in the tests but I could update those too.

lib/src/async/retry.dart
natebosch: Fixed the missing assignment to lastCaught. I was mistaken about the strength of my test for getting back the real exception - now updated to catch this mistake.

Good call on distinguishing against the task() throwing a TimeoutException. Updated to use onTimeout.

test/async/retry_test.dart
natebosch: I tried rewriting the tests using FakeAsync, and I don't think it buys us much. The crucial pieces are calls to Future#timeout and those don't take a Clock so we can't force them to trigger or to wait.

natebosch:
@jtmcdole any further thoughts on the interface?

jtmcdole:
Sorry it took me a while to look at. I like that it is a simple API.

PR #304, opened 2016-09-21
Version 0.23.0

cbracken:


lib/strings.dart
```diff
@@ -46,11 +46,13 @@ String flip(String s) {
 /**
  * If [s] is [null] returns empty string, otherwise returns [s] as is.
  */
+@Deprecated('Will be removed in 0.23.0')
```
yjbanov: Let's suggest an alternative in the deprecation message.

lib/strings.dart
```diff
@@ -46,11 +46,13 @@ String flip(String s) {
...
 String nullToEmpty(String s) => s == null ? '' : s;
 
 /**
  * If [s] is an empty string returns [null], otherwise returns [s] as is.
  */
+@Deprecated('Will be removed in 0.23.0')
```
yjbanov: ditto

lib/strings.dart
cbracken: Done.

lib/strings.dart
cbracken: Done.

xster:
The alternative for emptyToNull is still quite verbose no?
Consider the string `someList.where(...).map(...).join(...)`. It's hard to code golf it to one line of non-empty otherwise.

Issue #310, opened 2016-10-31
DDC error: Invalid override. The type of '_EqualityGroupMatcher.matches' ('(Map<String, List<dynamic>>, Map<dynamic, dynamic>) → bool') isn't a subtype of 'Matcher.matches' ('(dynamic, Map<dynamic, dynamic>) → bool'). (package:quiver/testing/src/equality/equality.dart, line 67, col 3)


zoechi:
```
 (master) $ bazelify serve
Building server via bazel...

WARNING: Sandboxed execution is not supported on your system and thus hermeticity of actions cannot be guaranteed. See http://bazel.io/docs/bazel-user-manual.html#sandboxing for more information. You can turn off this warning via --ignore_unsupported_sandboxing.
____Loading package: 
____Loading package: @bazel_tools//tools/cpp
____Loading package: @local_jdk//
____Loading package: @local_config_cc//
____Loading complete.  Analyzing...
____Loading package: @angular2//
____Loading package: @barback//
____Loading package: @glob//
____Loading package: @csslib//
____Loading package: @crypto//
____Loading package: @convert//
____Loading package: @stream_channel//
____Loading package: @io_bazel_rules_dart//vendor/path
____Loading package: @vendor_logging//
____Loading package: @io_bazel_rules_dart//vendor/shelf
____Loading package: @vendor_collection//
____Loading package: @io_bazel_rules_dart//vendor/typed_data
____Found 1 target...
____Building...
____[13 / 26] Summarizing Dart library @collection//:collection (strong mode)
____[42 / 72] Writing file bwu_ng_mdl_worker_args
____[125 / 211] Writing file bwu_ng_mdl_ddc_args
____[134 / 211] Compiling @intl//:intl with ddc
____[140 / 211] Compiling @yaml//:yaml with ddc
____[142 / 211] Compiling @async//:async with ddc
____[162 / 211] Summarizing Dart library @yaml//:yaml (strong mode)
____[165 / 211] Building Dart VM snapshot @io_bazel_rules_dart//dart/tools/ddc_html_generator:ddc_html_generator [for host]
____[166 / 211] Building Dart VM snapshot //:main_ddc_serve
ERROR: /home/zoechi/.cache/bazel/_bazel_zoechi/3e443aaa6a6cf1824cc4980f306f0a94/external/quiver/BUILD:14:1: Compiling @quiver//:quiver with ddc failed: Worker process sent response with exit code: 1..
[...]

PR #311, opened 2016-10-31
Strong mode fixes for EqualityGroupMatcher

cbracken:
Fix an inconsistent @override.


cbracken:
Fixes #310

/cc @yjbanov

PR #317, opened 2016-11-03
Add CountdownTimer.increment read-only property

cbracken:


cbracken:
HE'S ALIIIIVE! Thanks!

lib/src/async/countdown_timer.dart
```diff
@@ -24,8 +24,9 @@ class CountdownTimer extends Stream<CountdownTimer> {
   static const _THRESHOLD_MS = 4;
 
   final Duration _duration;
-  final Duration _increment;
   final Stopwatch _stopwatch;
+
+  final Duration increment;
```
yjbanov: Dart docs? Sorry for jumping in too late.

cbracken: PR #318

PR #319, opened 2016-11-03
Replace unnecessary use of Pair with containsPair matcher

cbracken:


test/collection/multimap_test.dart
```diff
@@ -96,12 +96,9 @@ void main() {
     });
 
     test('forEach should iterate over all key-value pairs', () {
```
yjbanov: The test is testing `forEach` explicitly. With this change it's no longer clear if `forEach` is exercised at all.

PR #321, opened 2016-11-04
Deprecate FutureGroup

cbracken:
Users should migrate to FutureGroup in package:async instead. Note that
that version requires a call to close() before it will complete when the
count of pending tasks reaches 0. This allows for use-cases where tasks
continue to be added after all existing tasks have completed.


cbracken:
/cc @yjbanov 

This replaces #263. Main differences are that we're not stipulating a removal release, since 100% of the use of `FutureGroup` internally is from Quiver and this will take a migration. Also provides a bit more detailed/objective commentary on the difference between the two.

Issue #122, opened 2014-04-10
deprecate strings repeat given Dart 1.3 String.* op?


chalin:
Might want to deprecate string repeat given that Dart 1.3 has introduced the `*` operator in `String` for this purpose.

cbracken:
Fixed by #323

PR #323, opened 2016-11-04
Deprecate repeat()

cbracken:
Users should migrate to the * operator on the String class.


cbracken:
/cc @yjbanov 

Fixes #122

PR #324, opened 2016-11-04
Rename flip() to reverse()

cbracken:
Deprecate flip() for one release, then remove.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
      pending tasks drops to 0.
    * Deprecated: `repeat` in the `strings` library. Use the `*` operator on
      the String class.
+   * Deprecated: in the strings library, `flip` has been renamed `reverse`.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+#### Unreleased
+   * Strings: Rename `flip` to `reverse`. `flip` is deprecated and will be
+     removed in one release.
```
yjbanov: "one"? Did you mean "1.0"? Or "a future"?

yjbanov: Also, this might technically be a breaking change since we could create conflicts with `reverse`s from other libraries.

cbracken: I mean in the current version++. Reworded to 'the next release'. 

In theory every addition of a new top-level identifier is a potentially breaking change. Not sure if it's worth calling them out as such though. Added a bit more detail to the changelog.

PR #325, opened 2016-11-04
Make use of async-await in forEachAsync

cbracken:


lib/src/async/iteration.dart
```diff
@@ -44,12 +47,12 @@ Future _doWhileAsync(
...
-Future _reduceAsync(Iterator iterator, currentValue, AsyncCombiner combine) {
+Future _reduceAsync(Iterator iterator, current, AsyncCombiner combine) async {
   if (iterator.moveNext()) {
-    return combine(currentValue, iterator.current)
-        .then((result) => _reduceAsync(iterator, result, combine));
+    var result = await combine(current, iterator.current);
+    return _reduceAsync(iterator, result, combine);
   }
-  return new Future.value(currentValue);
+  return new Future.value(current);
```
yjbanov: You should be able to just `return current;`

cbracken: Oops! How embarrassing! Done.

PR #329, opened 2016-12-10
Update changelog with recent changes

cbracken:
SDK constraint increased to 1.21.0 to support generic functions.
Iterables: enumerate is now generic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 #### Unreleased
-   * BREAKING CHANGE: minimum SDK constraint increased to 1.9.0. This allows
-     use of async-await in Quiver.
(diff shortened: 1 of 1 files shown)


yjbanov:
Could this be a little too aggressive? Maybe we should first move to the SDK version that supports the temporary generic syntax, then, say, end of January, move to the final syntax?

eukreign:
If users have quiver pegged at a specific version then if you release a new version it will not affect them.

If they don't have quiver pegged to a specific version then they probably want the latest and will also have the latest Dart version.

By releasing now users can decide for themselves if they want to upgrade today or end of January via their pubspec.yaml.

cbracken:
I think it's alright to merge now. The SDK was rolled last week and as @eukreign says, developers always have the option of narrowing their deps. Pub only upgrades when it's asked to or when its constraint solver tells it it needs to.

Issue #308, opened 2016-10-07
StreamBuffer refactor...


jtmcdole:
I'm not to happy with StreamBuffer at the moment. I thought it would be nice to have it handle both the cases of `Stream<List<T>>` and `Stream<T>` - Sockets give you the `List<int>` and other processes could generate a monotonic stream of events... all of which stream buffer neatly handled in `read()`.
#296 Changed it to `StreamConsumer<dynamic>` to remove a cast, however that makes other projects I work on have analyzer errors when trying to pipe a socket to the buffer.

What is the best way to make everyone happy?

``` dart
StreamBuffer<T> implements StreamConsumer<List<T>>
or
StreamBuffer<T> implements StreamConsumer<T>
```

`StreamConsumer<List<T>>` requires any monotonic controller to add a list of one element to the stream.
`StreamConsumer<T>` requires anyone using this as `StreamBuffer<int>` to update to `StreamBuffer<List<int>>`.

PR #334, opened 2016-12-14
StreamBuffer<T> now implements StreamConsumer<T>

cbracken:
Previously, it implemented StreamConsumer<T|List<T>>. For strong-mode
compatibility, we no longer support the union type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
    * BREAKING CHANGE: minimum SDK constraint increased to 1.21.0. This allows
      use of async-await and generic function in Quiver.
    * BREAKING CHANGE: eliminated deprecated `FakeTimer`.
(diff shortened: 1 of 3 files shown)


cbracken:
This gets the bots green again.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
    * BREAKING CHANGE: minimum SDK constraint increased to 1.21.0. This allows
      use of async-await and generic function in Quiver.
    * BREAKING CHANGE: eliminated deprecated `FakeTimer`.
+   * BREAKING CHANGE: `StreamBuffer<T>` now implements `StreamConsumer<T>` as
```
jtmcdole: This is less of a breaking change and more of a "return to normal" that was prior to #296

cbracken: UNBREAKING CHANGE!

cbracken:
Closes #308

PR #335, opened 2016-12-14
Enforce strong mode analysis in Travis

cbracken:


cbracken:
This gets the bots green again.

PR #341, opened 2017-02-16
added #indexOf to collection.dart

willyp564:
Here is an example use case: cs/piper///depot/google3/ads/awapps2/shared/ad_group_criterion/lib/ad_group_criterion_mutate_helper.dart?l=38

There are many others.

### DIFF
--- a/lib/collection.dart
+++ b/lib/collection.dart
@@ -78,13 +78,22 @@ bool setsEqual(Set a, Set b) {
   return a.containsAll(b);
 }
 
-/// Returns the index of the first item in [list] where [matcher] evaluates to
-/// true.
+/// Returns the index of the first item in [elements] where [predicate]
+/// evaluates to true.
 ///
-/// Returns -1 if there are no items where [matcher] evaluates to true.
-int indexOf<T>(List<T> list, bool matcher(T item)) {
-  for (var i = 0; i < list.length; i++) {
-    if (matcher(list[i])) return i;
+/// Returns -1 if there are no items where [predicate] evaluates to true.
+int indexOf<T>(Iterable<T> elements, bool predicate(T element)) {
+  if (elements is List) {
+    for (int i = 0; i < elements.length; i++) {
+      if (predicate(elements[i])) return i;
+    }
+    return -1;
+  }
+
+  int i = 0;
+  for (T element in elements) {
+    if (predicate(element)) return i;
+    i++;
   }
   return -1;
 }
--- a/lib/collection.dart
+++ b/lib/collection.dart
@@ -77,3 +77,14 @@ bool setsEqual(Set a, Set b) {
 
   return a.containsAll(b);
 }
+
+/// Returns the index of the first item in [list] where [matcher] evaluates to
+/// true.
+///
+/// Returns -1 if there are no items where [matcher] evaluates to true.
+int indexOf<T>(List<T> list, bool matcher(T item)) {
+  for (var i = 0; i < list.length; i++) {
+    if (matcher(list[i])) return i;
+  }
+  return -1;
+}
--- a/test/collection/collection_test.dart
+++ b/test/collection/collection_test.dart
@@ -69,4 +69,17 @@ main() {
       expect(setsEqual(new Set(), new Set.from([1])), isFalse);
     });
   });
+
+  group('indexOf', () {
+    test('returns the first matching index', () {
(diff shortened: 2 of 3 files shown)


lib/collection.dart
```diff
@@ -77,3 +77,14 @@ bool setsEqual(Set a, Set b) {
 
   return a.containsAll(b);
 }
+
+/// Returns the index of the first item in [list] where [matcher] evaluates to
+/// true.
+///
+/// Returns -1 if there are no items where [matcher] evaluates to true.
+int indexOf<T>(List<T> list, bool matcher(T item)) {
```
yjbanov: I think a more common name for the callback is `predicate`. `matcher` sounds like something from `package:matcher`. Similarly, a more common name for items in a collection is `element`.

yjbanov: I'd imagine this utility could be useful for any `Iterable`. Are we worried about performance? If so, we could do:

```dart
int indexOf<T>(Iterable<T> elements, bool predicate(T element)) {
  if (elements is List) {
    for (int i = 0; i < element.length; i++) {
      ...
    }
    return -1;
  }

  for (T element in elements) {
    ...
  }
  return -1;
}
```

willyp564:
yjbanov - I agree this could be useful for all Iterables. I also went ahead with your suggestion to increase performance.

willyp564:
Also, I believe cbracken told me the integration tests were failing because of something on your end. Although I could have misunderstood. Let me know if I need to take action on this.

Thanks!

cbracken:
On the integration tests: yep -- the issue is with `FakeAsync` tests. `package:test` v0.12.19 spins up a microtask in `expect()` which broke `FakeAsync` tests that expect a certain number of microtasks. Fixed in #342.

cbracken:
lgtm

test/collection/collection_test.dart
```diff
@@ -69,4 +69,17 @@ main() {
...
+  group('indexOf', () {
+    test('returns the first matching index', () {
+      expect(indexOf<int>([1, 12, 19, 20, 24], (n) => n % 2 == 0), 1);
+      expect(indexOf<String>(['a', 'b', 'a'], (s) => s == 'a'), 0);
+    });
+
+    test('returns -1 when there is no match', () {
+      expect(indexOf<int>([1, 3, 7], (n) => n % 2 == 0), -1);
+      expect(indexOf<String>(['a', 'b'], (s) => s == 'e'), -1);
+      expect(indexOf<bool>([], (n) => true), -1);
```
cbracken: nit: `(_) => true`

lib/collection.dart
```diff
@@ -77,3 +77,23 @@ bool setsEqual(Set a, Set b) {
...
 }
+
+/// Returns the index of the first item in [elements] where [predicate]
+/// evaluates to true.
+///
+/// Returns -1 if there are no items where [predicate] evaluates to true.
+int indexOf<T>(Iterable<T> elements, bool predicate(T element)) {
+  if (elements is List) {
+    for (int i = 0; i < elements.length; i++) {
+      if (predicate(elements[i])) return i;
```
cbracken: Iterable doesn't have `operator[]`, so you'll need a cast. I'd extract a local.
```dart
if (elements is List) {
  var elementList = elements as List;
  for (int i = 0; i < elementList.length; i++) {
    if (predicate(elements[i])) return i;
  }
}

...
```

Alternatively, stick with the original code. @yjbanov thoughts?

cbracken: Iterable doesn't have `operator[]`, so you'll need a cast. I'd extract a local.
```dart
if (elements is List) {
  var elementList = elements as List;
  for (int i = 0; i < elementList.length; i++) {
    if (predicate(elements[i])) return i;
  }
}

...
```

Alternatively, stick with the original code. @yjbanov thoughts?

yjbanov: I believe Dart does the type promotion when it sees an `is` check, so I think this should be strong-mode clean.

cbracken: The analyzer has had different behaviour over time with that. A manual run of the analyzer results in:
```
[error] The operator '[]' isn't defined for the class 'Iterable<T>'. (/home/travis/build/google/quiver-dart/lib/collection.dart, line 88, col 29)
```

cbracken: Alright changing to `List<T>` actually fixes the warning/error.

cbracken:
Alright manually patched up and merged in fcabebed1fd2f021ac328d9507b3a1447ba8cc21.

PR #349, opened 2017-04-06
Adding robust string split function

cjwawrzonek:
-Added robust string split function to strings.dart
-Added unit tests to strings_test.dart
-Modified docs

This split function includes a common use case of cleaning white space and removing empty splits from the returned list of Strings.


cjwawrzonek:
Unmerged. Reasons: use case too narrow, does not internationalize well.

PR #353, opened 2017-06-07
Improve reverse of string.

lrhn:
The existing implementation of reverse was quadratic in the string length.

It's still a very problematic method because it doesn't understand Unicode characters, only code points, so it will mis-attribute combining marks in a reversed string, e.g., reverse("niño") would become "õnin".

### DIFF
--- a/lib/strings.dart
+++ b/lib/strings.dart
@@ -28,9 +28,9 @@ bool isNotEmpty(String s) => s != null && s.isNotEmpty;
 String reverse(String s) {
   if (s == null || s == '') return s;
   StringBuffer sb = new StringBuffer();
-  var runes = s.runes;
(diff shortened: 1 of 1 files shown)


jtmcdole:
This looks like it still writes "õnin" if I put it in dartpad...

jtmcdole:
> Failed dartfmt check: run dartfmt -w lib/ test/

Just run dartfmt again and it's LGTM by me.

lrhn:
Yes, it doesn't fix the inherent problem in the function (which is that it doesn't work for real-life text strings).
That would require a way to recognize grapheme clusters, not just code units.

cbracken:
Thanks! Filed #354 for the composed character ranges bug.

PR #356, opened 2017-06-19
add a waitGroup async function

xster:


xster:
Linter wasn't me :) I think

PR #363, opened 2017-09-27
Remove deprecated strings.repeat

cbracken:
repeat was deprecated in f60deadf89036a9e52a23e29a78f20218951f07a.
Callers should use the '*' operator on String instead.


cbracken:
Sadly github doesn't support diffbasing, but I'd like to avoid a merge conflict on rebase ;-)

Ignore the first commit, which is just #362.

Issue #354, opened 2017-06-07
Strings reverse doesn't handle unicode composed characters


cbracken:
Example:
```
reverse('\u{0065}\u{0301}cole');
```
should return `elocé`, but instead returns `eloće`.

lrhn:
This invites the question of whether reversing a string is ever a good idea.

If the string contains *text* (words, phrases, meaning), reversing it is meaningless. If you need to read it right-to-left, Unicode has markers for that (U+200F RIGHT-TO-LEFT MARK), you don't need to reverse the sequence of code points physically

The case where reversing a string might make sense is when it doesn't contain text, just symbols. E.g., you want to to make a fancy ASCII header, so you do:
```dart
header(text, banner) => banner + text + reverse(banner);
print("Hello world!", "--==## "); // prints  "--==## Hello world! ##==--"
```
That requires you to know that the symbols are actually simple. Reversing modern emojis suffers the same problem as the accented text, you might switch the skin color to the wrong face.

Basically, I think `reverse(String)` is just enabling bad behavior, so you could deprecate it instead of fixing it. 

Also, fixing is *hard*. You need to recognize Unicode combining marks. And that's just the beginning.
The zero-width-joiner might combine other characters that are not combining marks. And some characters are combined just by juxtaposition, like regional indicator symbols (we don't want the flag for the United States becoming the flag of the Soviet Union when reversed). I doubt quiver wants to include a Unicode library - the size of the tables can be prohibitive.

All in all, Unicode is *not* built for reversal. If you keep the function, just document that it's only guaranteed to work for ASCII, it's the only way to stay sane :)

cbracken:
@lrhn all good points. A lot of what's in the Quiver strings library falls into the naughty list, though the limitations are mostly documented -- e.g. `isDigit` and `isWhitespace`. Unlike those two however, there's not a ton of use for `reverse` aside from novelty applications. I'm tempted to deprecate and delete it.

`loop` falls into a similar category though its parameters make the potential for badness at least a little clearer -- or at least not particularly worse than `substring` -- but again it feels like basically a novelty function.

PR #368, opened 2017-10-13
Deprecate Strings.reverse

cbracken:
Without full unicode support for composed character sequences,
sequences inclduing zero-width joiners, etc., this function is unsafe
for general use.

No replacement is provided since, aside from novelty applications, this
function has relatively few uses.


cbracken:
See discussion on #354.

Issue #124, opened 2014-04-11
Deprecate/remove {Create,create}[Periodic]Timer and FakeTimer


seaneagan:
They are all superceded by #98.

PR #369, opened 2017-10-13
Deprecate createTimer, createTimerPeriodic

cbracken:
These were originally authored in d5a10fec to support FakeTimer, which
was deprecated in be8d3772 and removed in 36cf71b5. FakeTimer was
superseded by FakeAsync, which provides a more comprehensive solution to
the problem these were originally written to solve.


cbracken:
Fixes #124.

PR #359, opened 2017-07-24
Change multimap to be backed by a LinkedHashMap.

har79:
To fix https://github.com/google/quiver-dart/issues/358

PR #370, opened 2017-10-14
Update changelog to note Multimap is now backed by LinkedHashMap

cbracken:


cbracken:
Reflects change introduced in #359.

PR #380, opened 2017-10-30
Fix missing variable name in fake_async

### LINKED ISSUE #379: _drainTimersWhile signature is missing variable name in predicate signature
The `predicate` argument to `_drainTimersWhile` in `fake_async.dart` uses `_FakeTimer` as a variable name where it is almost certainly intended to be used as the argument type.

This results in unnecessary `dynamic as bottom` hints at use sites.

a14n:
Fixes #379


cbracken:
lgtm

Issue #378, opened 2017-10-29
Consider making _daysInMonth and _isLeapYear public


jolleekin:
When creating a date picker Polymer element, a date picker Angular component, or a subscription management system, developers will likely need `_daysInMonth` and `_isLeapYear`. If these two functions are made public, developers only need to import this package instead of defining their own.

cbracken:
Seems like totally reasonable things for Quiver to expose. I can imagine that long term, we might want to expose a clean set of time APIs, but we can always expose these as top-level functions, then if we ever get around to building something better, add the new APIs and deprecate these ones.

@yjbanov @a14n thoughts?

cbracken:
Landed in #381, will be released in v0.26.0, in #382.

cbracken:
And 0.26.0 is released on pub! @jolleekin Thanks for suggesting this!

PR #381, opened 2017-11-01
Add isLeapYear, daysInMonth, clampDayOfMonth

cbracken:
This change extracts these functions to a new file, util.dart, and makes
them public.

### DIFF
--- a/lib/src/time/util.dart
+++ b/lib/src/time/util.dart
@@ -1,0 +1,58 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of quiver.time;
+
+/// Days in a month. This array uses 1-based month numbers, i.e. January is
+/// the 1-st element in the array, not the 0-th.
+const _daysInMonth = const [0, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+
+/// Returns the number of days in the specified month.
+///
+/// This function assumes the use of the Gregorian calendar or the proleptic
+/// Gregorian calendar.
+int daysInMonth(int year, int month) =>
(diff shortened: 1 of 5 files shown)


cbracken:
Fixes #378.

lib/src/time/util.dart
```diff
@@ -0,0 +1,59 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of quiver.time;
+
+/// Days in a month. This array uses 1-based month numbers, i.e. January is
+/// the 1-st element in the array, not the 0-th.
+const _DAYS_IN_MONTH = const [
```
yjbanov: lowerCamelCase

lib/src/time/util.dart
```diff
@@ -0,0 +1,59 @@
...
+  30,
+  31,
+  31,
+  30,
+  31,
+  30,
+  31
+];
+
+int daysInMonth(int year, int month) =>
```
yjbanov: dartdocs

lib/src/time/util.dart
```diff
@@ -0,0 +1,59 @@
...
+    (month == DateTime.FEBRUARY && isLeapYear(year))
+        ? 29
+        : _DAYS_IN_MONTH[month];
+
+bool isLeapYear(int year) =>
```
yjbanov: dartdocs

lib/src/time/util.dart
```diff
@@ -0,0 +1,59 @@
...
+///
+/// February 31, 2013 => February 28, 2013
+///
+/// When jumping from month to month or from leap year to common year we may
+/// end up in a month that has fewer days than the month we are jumping from.
+/// In that case it is impossible to preserve the exact date. So we "clamp" the
+/// date value to fit within the month. For example, jumping from March 31 one
+/// month back takes us to February 28 (or 29 during a leap year), as February
+/// doesn't have 31-st date.
+int clampDate(int date, int year, int month) =>
```
yjbanov: Is this a good name? `clampDateToMonth`?

yjbanov: Should parameters be named? I hear `DateTime` is changing its signature to name _its_ parameters.

yjbanov:
Also, what do you think of moving these into a new library named `calendar`? It's weird from the discoverability standpoint to put them in `time.dart`.

yjbanov:
lgtm

lib/src/time/util.dart
cbracken: Done - looks like we almost certainly depend on `meta` via `matcher` anyway` :)

cbracken: Done: `clampDayOfMonth`

Issue #385, opened 2017-11-16
LinkedLruHashMap: Bad state when remove() removes the last entry


hugovangalen:
My own code crashed in the method `_removeLru` with a null pointer access when calling `_tail.next`.

After debugging I discovered that the cause seems to be in the `remove()` method. 

There exists a check whether the removed `entry` is equal to `_head`, or it is equal to `_tail`. 

If it happens to be the last item that gets removed, that is not handled correctly. Right now it branches into the `entry == _head`, resets _head, but leaves _tail in its old value, which is wrong. 

If my understanding is correct, in that case both `_tail` and `_head` need to be reset to null. Adding a test for `entry == _head && entry == _tail` to reset the `_tail` and `_head` to `null` appears to fix the problem:

```
  @override
  V remove(Object key) {
    final entry = _entries.remove(key);
    if (entry != null) {
      if (entry == _head && entry == _tail) {
        _head = null;
        _tail = null;
      } else if (entry == _head) {
        _head = _head.next;
      } else if (entry == _tail) {
        _tail.previous.next = null;
        _tail = _tail.previous;
      } else {
        entry.previous.next = entry.next;
      }

      return entry.value;
    }
    return null;
  }
```

cbracken:
Fixed by #386 and released in v0.26.1. Thanks for diagnosing and reporting this!

hugovangalen:
@cbracken No problem. I'd like to stress that there are still issues with dangling references. See ticket #412.

PR #386, opened 2017-11-16
Fix bad tail state in LRU map on removal of last entry

cbracken:
On removal of the last entry from an LRU map, the entry is both _head
and _tail. The existing remove() implementation matched it against
_head, which was then updated to _head.next (null), but _tail was left
unmodified in a bad state.

Since _tail is not null, the next insertion into the newly-empty map
don't reset it. When the map size eventually exceeds the maximum size,
_removeLru() is called to evict _tail, resulting in a null-pointer
exception as _removeLru() attempts to get _tail.previous (null) and then
set .next on that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+#### 0.27.0-dev
+
+   * Fix: when removing the last item, `LinkedLruHashMap` was put into a state
+     such that the next cache eviction could cause a null-pointer exception.
+     Issue [#385](https://github.com/google/quiver-dart/issues/385).
+
 #### 0.26.0 - 2017-11-01
    * BREAKING CHANGE: eliminated deprecated `flip`. Replaced by `reverse` in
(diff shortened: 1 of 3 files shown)


cbracken:
Fixes #385.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+#### 0.27.0-dev
```
tvolkert: If you plan on releasing this now, then bump the pubspec version as well

cbracken: I'm actually planning to do a trick here and release v0.26.1 since all changes since 0.26.0 are backwards-compatible. Keeping this as-is though since the current pubspec version is 0.27.0-dev and I've got one more update to land since we set the dev version. 

Typically we update this value in a release commit that updates the pubspec, the readme and the changelog atomically.

cbracken:
Released in v0.26.1

PR #393, opened 2017-11-16
Use the expanded test reporter

cbracken:
This ensures we emit one line per test as opposed to overwriting with
each pass.


yjbanov:
lgtm

Issue #258, opened 2015-05-24
FakeAsync: Timers are drained in mostly-recently-scheduled (incorrect) order


matanlurey:
Open Chrome, try the following:

```js
function tryOrder() {
  setTimeout(function () {
    console.log('Timer 1');
  });
  setTimeout(function () {
    console.log('Timer 2');
  });
}
```

Run `tryOrder()`

Observe:
```
Timer 1
Timer 2
```

However, FakeAsync does the following:

```dart
void tryOrder() {
  new FakeAsync().run((fakeAsync) {
    Timer.run(() => print('Timer 1'));
    Timer.run(() => print('Timer 2'));
    fakeAsync.elapse(Duration.ZERO);
  });
}

void main() {
  tryOrder();
}
```

Observe:
```
Timer 2
Timer 1
```

PR #265, opened 2015-08-13
Use FIFO timer ordering in FakeAsync (fixes #258)

alechenninger:

PR #395, opened 2017-11-16
Use FIFO timer ordering in FakeAsync

cbracken:
As noted in https://github.com/google/quiver-dart/issues/258, the
current ordering of _getNextTimer() in _FakeAsync is inconsistent with
the native Dart event loop.

In the process of fixing that, I had to fix a few other idiosyncrasies
that were revealed from the existing unit tests which relied on the
incorrect behavior. For instance, if you had a periodic timer occur at
the same time as a delayed timer, whether or not both of them were
"flushed" in flushTimers depended on the order with which they were
queued. Now it does not.

### DIFF
--- a/lib/testing/src/async/fake_async.dart
+++ b/lib/testing/src/async/fake_async.dart
@@ -170,8 +170,10 @@ class _FakeAsync implements FakeAsync {
       if (flushPeriodicTimers) {
         return _timers.isNotEmpty;
       } else {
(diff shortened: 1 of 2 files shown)


cbracken:
Note this is a rebased pull of #265 for testing purposes. Once Travis has passed, I'll manually merge that one.

cbracken:
Looks good. Closing here and manually merging.

Issue #255, opened 2015-05-17
Cache will make multiple requests for same key 


hangstrap:
Assume that the isAbsent function return a future. 
A call is make to access key "foo", this results in a cache miss and isAbsend it called. 
Before isAbsent returns another call is made it access key "foo", and once more isAbsent is called.

This is a bug as  isAbsent has been called twice for the same key. The second call should be 'blocked' until the first isAbsent returns.

```
  test( "test", () async{

    int count=0;
    Future<String> loader (String key){
      count ++;
      print( "inside");
      return new Future.delayed( new Duration( seconds:1), ()=>"Test");
    }

    MapCache<String, String> cache = new MapCache.lru();
    FutureGroup f = new FutureGroup();
    f.add( cache.get( "test", ifAbsent:loader));
    f.add( cache.get( "test", ifAbsent:loader));

    await f;
    expect( count, equals(1));
  });
```

matanlurey:
@cbracken could you assign this to me? I'm observing the same bug internally, will fix this week.

adamlofts:
Do you have an internal fix you can pick and apply to this? If not this is a bug I would like to fix can can work on next week.

PR #398, opened 2017-11-20
MapCache: Prevent concurrent loads of the same key (#255)

adamlofts:
Keep the set of outstanding calls to ifAbsent so that we can return the existing
Future if asked for the same key.

Fix: #255

### DIFF
--- a/lib/src/cache/map_cache.dart
+++ b/lib/src/cache/map_cache.dart
@@ -17,6 +17,8 @@ part of quiver.cache;
 /// A [Cache] that's backed by a [Map].
 class MapCache<K, V> implements Cache<K, V> {
   final Map<K, V> _map;
+  // Map of outstanding ifAbsent calls used to prevent concurrent loads of the same key.
+  final Map<K, FutureOr<V>> _outstanding = {};
 
   /// Creates a new [MapCache], optionally using [map] as the backing [Map].
   MapCache({Map<K, V> map}) : _map = map != null ? map : new HashMap<K, V>();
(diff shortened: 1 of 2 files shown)


lib/src/cache/map_cache.dart
```diff
@@ -17,6 +17,8 @@ part of quiver.cache;
 /// A [Cache] that's backed by a [Map].
 class MapCache<K, V> implements Cache<K, V> {
   final Map<K, V> _map;
+  // Map of outstanding ifAbsent calls used to prevent concurrent loads of the same key.
```
a14n: Use `///` instead of `//` ?

lib/src/cache/map_cache.dart
```diff
@@ -17,6 +17,8 @@ part of quiver.cache;
...
+  final Map<K, FutureOr<V>> _outstanding = {};
```
a14n: `final _outstanding = <K, FutureOr<V>>{}` otherwise `_outstanding` will be `Map<dynamic, dynamic>` at runtime.

test/cache/map_cache_test.dart
```diff
@@ -64,5 +64,21 @@ main() {
         expect(value, 'foofoo');
       });
     });
+
+    test("should not make multiple requests for the same key", () async {
+      int count = 0;
+
+      Future<String> loader(String key) {
+        count += 1;
+        return new Future.delayed(new Duration(milliseconds: 1), () => "test");
```
a14n: `const Duration`

cbracken:
lgtm

cbracken:
Thanks for this! Was just asking around about this issue last week :)

adamlofts:
👍 thanks for merging.

PR #406, opened 2017-12-22
Update with Dart 2.0 core library unimplemented implementations

srawlins:
Updated the collection classes and the timer class with new Dart 2.0 default implementations, from [this gist](https://gist.github.com/lrhn/b99fe5f73a10b3b1a91579853056d08b).

Once the base collection classes have docs and implementations, we can flesh these out. This PR just allows a user to depend on quiver while using Dart 2.0 once the collection classes have new methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 #### 0.27.0-dev
 
+   * BREAKING CHANGE: all classes that implement `Iterable`, `List`, `Map`,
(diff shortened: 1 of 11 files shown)


srawlins:
Done! Ship it!

srawlins:
@cbracken To answer your question: To the best of my knowledge, this fixes all of the planned Dart 2.0 core lib changes that would affect quiver.

PR #414, opened 2018-02-26
Remove usage of Maps.mapToString()

srawlins:
`Maps.mapToString()` is going to be removed by Dart 2.0. It will be moved to `MapBase.mapToString()`. To avoid bumping the minimum SDK version, we can just copy the implementation (~30 lines) to this package.

(This is the same fix that was applied to collection, https://github.com/dart-lang/collection/pull/71.)


srawlins:
CC @cbracken

PR #419, opened 2018-03-05
Add Optional.transformNullable.

mfiels:
This lets callers chain maybe-present values through transformers with a nullable return value.

The existing Optional.transform throws an `ArgumentException` if the transformer returns `null`. Would be great to have an API that returns `absent()` if the transformer returns `null`.

I added this as a new function for backwards compatibility sake, but I'd argue that Optional.transform should have this behavior -- depends on what your policy is for breaking changes.

Inspiration taken from Java 8's [`Optional#map`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-):

>If a value is present, apply the provided mapping function to it, and if the result is non-null, return an Optional describing the result. Otherwise return an empty Optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+#### 2.0.1 - 2018-10-22
+
+   * New: Optional now includes `transformNullable` to pass maybe present
+     values through a transformer with a nullable return value.
+
 #### 2.0.0+1 - 2017-07-18
 
    * Updated Dart SDK constraint to >=2.0.0-dev.61 < 3.0.0.
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -90,6 +90,17 @@ class Optional<T> extends IterableBase<T> {
         : new Optional.of(transformer(_value));
   }
 
+  /// Transforms the Optional value.
+  ///
+  /// If the Optional is [absent()], returns [absent()] without applying the transformer.
+  ///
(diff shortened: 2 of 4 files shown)


mfiels:
@cbracken thoughts?

yjbanov:
@davidmorgan

davidmorgan:
quiver `Optional` was based on Guava `Optional`, which predates Java 8 `Optional`. This is an API difference between the two.

I don't think we should change to emulating a different `Optional`; it's not worth the pain.

Given that, `transformNullable` seems like a good compromise.

yjbanov:
@davidmorgan Is that an LGTM from you? :smile:

davidmorgan:
@yjbanov Well, I didn't look at the actual code :D ... now that I do, yes, LGTM.

yjbanov:
Cool! In that case let's add it to https://github.com/google/quiver-dart/blob/master/CHANGELOG.md and merge.

davidmorgan:
Sounds good. I don't have access to merge, can you? @yjbanov @mfiels

yjbanov:
@mfiels Could you please bump the version number to `2.0.1` and the new feature to `CHANGELOG.md`? I'll then merge and publish.

yjbanov:
I rebase-merged and kept the two commits. Thanks for contributing! I'll publish a new version shortly.

PR #420, opened 2018-03-05
Add tests of TreeSet removal operations

### LINKED ISSUE #417: Low test coverage for TreeSet
Test coverage isn't everything, but I'd expect a low level package like quiver to have near 100% coverage. Currently the file with the most _uncovered_ lines (113 lines) is [treeset.dart](https://coveralls.io/builds/15765008/source?filename=lib%2Fsrc%2Fcollection%2Ftreeset.dart). Coverage could be improved, especially among some of the balancing logic.

srawlins:
Should maybe fix #417.

### DIFF
--- a/test/collection/treeset_test.dart
+++ b/test/collection/treeset_test.dart
@@ -302,60 +302,60 @@ main() {
       });
 
       test("remove from tree", () {
-        tree = new TreeSet()..addAll([10,20,15,21,30,20]);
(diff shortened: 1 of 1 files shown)


srawlins:
Well, we can't see Coveralls results until this lands 🙁 so... I guess it should land on its own merits which are that it introduces tests for methods that had no tests previously.

Issue #418, opened 2018-03-02
Implement Dart 2.0 collection methods


srawlins:
I introduced a ton of [Unimplemented](https://github.com/google/quiver-dart/search?utf8=%E2%9C%93&q=Unimplemented&type=) stubs recently. Now that quiver has a minimum version constraint that guarantees that the new methods are available, we should implement them.

cbracken:
@srawlins I've x-ref'ed the fixes here. Is this now complete?

srawlins:
No sir! Still have 34 Unimplemented methods:

```shell
$ git grep Unimplemented | wc -l
34
```

alanrussian:
It was very surprising to me to see an UnimplementedError when I performed a normal (Dart 2) map operation. Hope this can get fixed soon.

srawlins:
Sorry about that. This is sort of a "Contributions welcome" situation. Or maybe the Quiver owners want to own this.

cbracken:
This is now down to 22 remaining. 

Working on knocking off the rest. Contributions are of course most welcome :)

cbracken:
Remainder:
* [ ] lib/src/collection/bimap.dart:107
* [ ] lib/src/collection/delegates/iterable.dart:35
* [ ] lib/src/collection/delegates/iterable.dart:126
* [ ] lib/src/collection/delegates/list.dart:54
* [ ] lib/src/collection/delegates/map.dart:43
* [ ] lib/src/collection/delegates/map.dart:51
* [ ] lib/src/collection/delegates/map.dart:66
* [ ] lib/src/collection/delegates/map.dart:89
* [ ] lib/src/collection/delegates/map.dart:103
* [ ] lib/src/collection/delegates/map.dart:109
* [ ] lib/src/collection/delegates/map.dart:115
* [ ] lib/src/collection/delegates/queue.dart:46
* [ ] lib/src/collection/delegates/set.dart:40
* [ ] lib/src/collection/lru_map.dart:91
* [ ] lib/src/collection/lru_map.dart:152
* [ ] lib/src/collection/multimap.dart:334
* [ ] lib/src/collection/multimap.dart:403
* [ ] lib/src/collection/multimap.dart:577
* [ ] lib/src/collection/multimap.dart:626
* [ ] lib/src/collection/multimap.dart:821
* [ ] lib/src/collection/treeset.dart:398
* [ ] lib/testing/src/async/fake_async.dart:303

HerrNiklasRaab:
[...]

PR #423, opened 2018-03-09
Add new Dart 2 Map methods to BiMap

srawlins:


lib/src/collection/bimap.dart
```diff
@@ -70,13 +70,18 @@ class HashBiMap<K, V> implements BiMap<K, V> {
   int get length => _map.length;
   Iterable<V> get values => _inverse.keys;
 
+  BiMap<V, K> get inverse {
+    if (_cached == null) {
+      _cached = new HashBiMap._from(_inverse, _map);
+    }
+    return _cached;
+  }
```
a14n: With null-aware assignment:

```dart
BiMap<V, K> get inverse => _cached ??= new HashBiMap._from(_inverse, _map);
```

lib/src/collection/bimap.dart
```diff
@@ -87,28 +92,18 @@ class HashBiMap<K, V> implements BiMap<K, V> {
...
-    throw new UnimplementedError("entries");
-  }
-
-  BiMap<V, K> get inverse {
-    if (_cached == null) {
-      _cached = new HashBiMap._from(_inverse, _map);
-    }
-    return _cached;
+  Iterable<MapEntry<K, V>> get entries {
+    return keys.map((K key) => new MapEntry<K, V>(key, this[key]));
```
a14n: Why not returning `_map.entries` ?

```dart
Iterable<MapEntry<K, V>> get entries => _map.entries;
```

lib/src/collection/bimap.dart
```diff
@@ -87,28 +92,18 @@ class HashBiMap<K, V> implements BiMap<K, V> {
...
-    // Change Object to MapEntry<K2, V2> when
-    // the MapEntry class has been added.
-    throw new UnimplementedError("map");
+  Map<K2, V2> map<K2, V2>(MapEntry<K2, V2> transform(K key, V value)) {
+    var result = <K2, V2>{};
+    for (var key in this.keys) {
+      var entry = transform(key, this[key]);
+      result[entry.key] = entry.value;
+    }
+    return result;
```
a14n: Why not returning `_map.map(transform)` ?

```dart
  Map<K2, V2> map<K2, V2>(MapEntry<K2, V2> transform(K key, V value)) {
    return _map.map(transform);
  }
```

srawlins: Done.

lib/src/collection/bimap.dart
srawlins: Done.

lib/src/collection/bimap.dart
srawlins: Love it. Done.

lib/src/collection/bimap.dart
```diff
@@ -70,13 +70,13 @@ class HashBiMap<K, V> implements BiMap<K, V> {
...
   @override
-  // TODO: Dart 2.0 requires this method to be implemented.
-  // ignore: override_on_non_overriding_method
-  void addEntries(Iterable<Object> entries) {
-    // Change Iterable<Object> to Iterable<MapEntry<K, V>> when
-    // the MapEntry class has been added.
-    throw new UnimplementedError("addEntries");
+  void addEntries(Iterable<MapEntry<K, V>> entries) {
+    for (var entry in entries) {
+      this[entry.key] = entry.value;
```
cbracken: Implement this using `_add` since it should apply the same invariants.

lib/src/collection/bimap.dart
```diff
@@ -87,29 +87,11 @@ class HashBiMap<K, V> implements BiMap<K, V> {
...
   @override
-  // TODO: Dart 2.0 requires this method to be implemented.
-  // ignore: override_on_non_overriding_method
-  Map<K2, V2> map<K2, V2>(Object transform(K key, V value)) {
-    // Change Object to MapEntry<K2, V2> when
-    // the MapEntry class has been added.
-    throw new UnimplementedError("map");
-  }
+  Map<K2, V2> map<K2, V2>(MapEntry<K2, V2> transform(K key, V value)) =>
+      _map.map(transform);
```
cbracken: Does `Map.map` document the expected return type? If it's expected to be a `BiMap`, we need to apply the invariants from `_add` (i.e. that not only must the keys be unique, but so must the values, since they're the keys of the inverse).

If it's not, then I think this is fine as-is.

lib/src/collection/bimap.dart
```diff
@@ -124,10 +106,9 @@ class HashBiMap<K, V> implements BiMap<K, V> {
   }
 
   @override
-  // TODO: Dart 2.0 requires this method to be implemented.
-  // ignore: override_on_non_overriding_method
   void removeWhere(bool test(K key, V value)) {
-    throw new UnimplementedError("removeWhere");
+    _map.removeWhere(test);
+    _inverse.removeWhere((v, k) => test(k, v));
```
cbracken: nit: Let's do this in the same order as `remove`... not that it really matters :)

lib/src/collection/bimap.dart
srawlins: [It doesn't document really.](https://api.dartlang.org/dev/2.0.0-dev.37.0/dart-core/Map/map.html)

My thought is it looks just like `List.map`, Set.map`, ...: you don't expect a List or a Set out the other side; you just expect an Iterable.

lib/src/collection/bimap.dart
srawlins: Done.

lib/src/collection/bimap.dart
srawlins: Done.

lib/src/collection/bimap.dart
cbracken: Cool. That sounds perfectly reasonable to me.

srawlins:
R rrr rrrrrrready to merge!

cbracken:
Partial fix to #418.

PR #424, opened 2018-03-09
Document #373 as breaking change

### LINKED ISSUE #404: #373 is an undocumented breaking change.
I had to revert https://github.com/google/quiver-dart/pull/373 to update Quiver internally.

Optional didn't use to check that `T` was equivalent. We will need a volunteer to update this in Google or this should be reverted.

srawlins:
Fixes #404


cbracken:
lgtm

Issue #399, opened 2017-12-07
#371 is an undocumented breaking change


matanlurey:
https://github.com/google/quiver-dart/pull/371

We can't land Quiver into Google anymore without patches. Consider reverting, or someone (i.e. not me) needs to spend the time updating Google before landing any new versions of Quiver (I'm handling the latest release, with a manual revert of #371).

a14n:
Sorry for the breakage. I didn't thought this PR could lead to problems. Could you show some examples where the new version of quiver causes problems?

matanlurey:
This causes a build failure:

```dart
abstract class Cache {
  Future<Null> clear();
}

class UserCache implements Cache {
  // ERROR: Expected Future<Null>.
  Future clear() => ...
}
```

cbracken:
Thanks for getting the bulk of the roll done @matanlurey. I'll get the remainder rolled internally in early Jan.

cbracken:
Documented in #441 on master and in eb2ad0c on the `dart_1` branch.

PR #441, opened 2018-03-24
Update 0.26.0 changelog with breaking change

cbracken:
This documents a breaking change that occurred in 5a9d955 which added
stronger types on `Cache` and `MapCache`.


cbracken:
Fixes #399.

PR #448, opened 2018-03-24
Run dartdevc-compiled web tests on Travis

cbracken:
Re-enables DDC web tests on Travis.

The build_runner package makes DDC testing quick and easy, but
indirectly depends on quiver, which introduces a circular dependency. To
avoid this, we patch the pubspec for the test run only.

This change also sets `env` to a single variable rather that a list of one var.

### DIFF
--- a/tool/travis/ddc_test.sh
+++ b/tool/travis/ddc_test.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


yjbanov:
I love this trick!

lgtm

tool/travis/ddc_test.sh
```diff
@@ -0,0 +1,45 @@
...
+cp -r . "$TEST_DIR"
+
+pushd "$TEST_DIR" > /dev/null 2>&1
+
+# Hack additional dev dependencies into pubspec.yaml
+#
+# We can't check these in because build_runner indirectly depends on quiver,
+# which introduces a circular dependency and makes pub upset.
+echo "Patching pubspec..."
+echo "  build_runner: ^0.7.11" >> pubspec.yaml
```
natebosch: we should be able to use `0.8.0`

cbracken: Done.

PR #451, opened 2018-03-26
Eliminate deprecated FutureGroup

cbracken:
FutureGroup was deprecated in v0.25.0. Users should migrate to
FutureGroup in package:async.


yjbanov:
lgtm

PR #460, opened 2018-05-30
Fix use of deprecated dart:io constant.

lrhn:
Update SDK dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 #### Master
 
+   * This version requires Dart SDK 2.0.0-dev.55 or later.
+
+#### 0.29.0+1 - 2018-03-29
+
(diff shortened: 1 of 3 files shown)


jtmcdole:
When is Dart 2.0 going to be released to stable?

cbracken:
LGTM. Thanks Lasse!

@jtmcdole my thought is we should probably do any last cleanup we want on both `master` (Dart 2) and `dart_1` branches, then publish 1.0.0 and 2.0.0-dev simultaneously. Today we still only publish `dart_1` to pub.dartlang.org, and those who want Dart 2 (e.g. in google3) grab it directly from git. Feature-wise, both branches are at parity so seems reasonable. There's still some work to do filling in some of the unimplemented methods on the collections classes.

Issue #465, opened 2018-06-29
partition: Isn't Dart2 compatible


matanlurey:
https://github.com/google/quiver-dart/blob/04ea48aa9d3af6af286d34948fbc5d5b2d1dfdc6/lib/src/iterables/partition.dart#L18-L20

yjbanov:
Fixed by https://github.com/google/quiver-dart/pull/488

PR #466, opened 2018-07-02
Converting the partition method to be generic, instead of always returning dynamic

lucyget:


lucyget:
@matanlurey I believe this should fix #465

matanlurey:
Apologies, I am OOO on vacation. Hopefully @cbracken or @yjbanov can take a look soon.

PR #471, opened 2018-07-18
misc: create v0.29.9 release to allow an easier transition to Dart 2

kevmoo:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+#### 0.29.9 - 2018-07-18
+
+   * Updated usage of Dart SDK constants for Dart v2.
+     Allows users of `pkg:quiver` to support Dart 2 gold without addressing
+     all of the breaking changes in `pkg:quiver` v2.
```
cbracken: quiver 2.0 is commit-for-commit identical to quiver 1.0 other than changes to update for Dart 2.

kevmoo:
> This violates semantic versioning, and likely requires further changes that will be incompatible with Dart 1.

How? Just trying to understand an example

> The correct fix is to release build_* packages with updated quiver dependencies.

Yes, but...

1)  I can't get a commit into `built_*` until Monday at the latest – and that's too late
2) Many more ecosystem packages use `quiver` – this allows us to minimize the number of blowups across the ecosystem when we roll out the const fix

I'm not saying this approach is ideal – far from it – but I do think it's *by far* the most pragmatic.

cbracken:
Wasn’t aware of that package. I’d be much more comfortable with that approach. On my phone and travelling out of the country. I could review but you’ll need someone to publish it. I think Jacob R has the upload bit set.

kevmoo:
Running with that now - https://github.com/google/quiver-dart/pull/472

Issue #461, opened 2018-06-13
MapCache: Can't clear a failed Future


creisman:
https://github.com/google/quiver-dart/blob/98dd92f4fd89fb8262a557db259e3bba7fde18cb/lib/src/cache/map_cache.dart

The value gets added to _outstanding, then is awaited. If the await throws an exception, it gets kicked out of the method and never clears the key from _outstanding. Subsequent calls will always return the failed Future. There's no way to evict a key from the _outstanding map so it's permanently broken for that key.

The [await](https://github.com/google/quiver-dart/blob/98dd92f4fd89fb8262a557db259e3bba7fde18cb/lib/src/cache/map_cache.dart#L44) either needs to be in a try/finally that removes it from _outstanding or invalidate needs to also try and remove keys from _outstanding.

creisman:
@adamlofts

yjbanov:
Fixed in https://github.com/google/quiver-dart/commit/252d1c71ac171d9044f145a301d19749b1cabba1

PR #473, opened 2018-07-19

deepjoy:

### DIFF
--- a/test/cache/map_cache_test.dart
+++ b/test/cache/map_cache_test.dart
@@ -81,5 +81,21 @@ main() {
 
       expect(count, equals(1));
     });
+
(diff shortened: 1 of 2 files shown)


test/cache/map_cache_test.dart
```diff
@@ -81,5 +81,21 @@ main() {
 
       expect(count, equals(1));
     });
+
+    test("should not cache a failed request", () async {
+      int count = 0;
+
+      Future<String> loader(String key) {
```
jacob314: nit you can write this a little more simply using async.

Future<String> loader(String key) async {
  count += 1;
  throw "Request failed";
}

PR #474, opened 2018-07-19

deepjoy:

### DIFF
--- a/test/cache/map_cache_test.dart
+++ b/test/cache/map_cache_test.dart
@@ -87,12 +87,14 @@ main() {
 
       Future<String> loader(String key) {
         count += 1;
-        return new Future.error("Request failed");
+        return new Future.error(new StateError("Request failed"));
       }
 
       await Future.wait(<Future>[
-        expectThrows(() => cache.get("test", ifAbsent: loader)),
-        expectThrows(() => cache.get("test", ifAbsent: loader)),
+        expect(() => cache.get("test", ifAbsent: loader),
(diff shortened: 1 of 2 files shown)


test/cache/map_cache_test.dart
```diff
@@ -81,5 +81,23 @@ main() {
...
+      }
+
+      await Future.wait(<Future>[
+        expect(() => cache.get("test", ifAbsent: loader),
+            throwsA(const TypeMatcher<StateError>())),
+        expect(() => cache.get("test", ifAbsent: loader),
+            throwsA(const TypeMatcher<StateError>())),
+      ]);
+
+      expect(count, equals(2));
```
a14n: add `expect(cache.get('test'), null)` ?

test/cache/map_cache_test.dart
```diff
@@ -81,5 +81,23 @@ main() {
...
       expect(count, equals(1));
     });
+
+    test("should not cache a failed request", () async {
+      int count = 0;
+
+      Future<String> loader(String key) {
+        count += 1;
+        return new Future.error(new StateError("Request failed"));
+      }
```
a14n: Alternatively you can use async function:

```dart
Future<String> loader(String key) async {
  count += 1;
  throw new StateError("Request failed");
}
```

yjbanov:
This has been superseded by https://github.com/google/quiver-dart/pull/491.

yjbanov:
Also, it seems @a14n's comments were addressed in the new PR.

PR #476, opened 2018-08-10
Fix MISSING_RETURN error in concat.dart

srawlins:
Having the return statement led... inference or something to think that this function could have return values, and then it didn't _always_ return.

### DIFF
--- a/lib/src/async/concat.dart
+++ b/lib/src/async/concat.dart
@@ -71,7 +71,7 @@ class _ConcatStream<T> extends Stream<T> {
     }, onResume: () {
       if (currentSubscription != null) currentSubscription.resume();
     }, onCancel: () {
-      if (currentSubscription != null) return currentSubscription.cancel();
+      if (currentSubscription != null) currentSubscription.cancel();
     });
 
     nextStream();
--- a/lib/src/async/concat.dart
+++ b/lib/src/async/concat.dart
@@ -71,8 +71,10 @@ class _ConcatStream<T> extends Stream<T> {
     }, onResume: () {
       if (currentSubscription != null) currentSubscription.resume();
     }, onCancel: () {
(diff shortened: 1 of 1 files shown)


srawlins:
CC @yjbanov

srawlins:
I have no write access here. @a14n mind re-trying Travis and/or merging? Thanks!

lib/src/async/concat.dart
```diff
@@ -71,7 +71,7 @@ class _ConcatStream<T> extends Stream<T> {
     }, onResume: () {
       if (currentSubscription != null) currentSubscription.resume();
     }, onCancel: () {
-      if (currentSubscription != null) return currentSubscription.cancel();
+      if (currentSubscription != null) currentSubscription.cancel();
```
yjbanov: I think this is wrong. `onCancel` returns `void` or `Future`. Removing `return` will drop the future returned by `cancel()` on the floow. The proper fix is probably to `return null` in the `else` branch.

lib/src/async/concat.dart
```diff
@@ -71,7 +71,7 @@ class _ConcatStream<T> extends Stream<T> {
     }, onResume: () {
       if (currentSubscription != null) currentSubscription.resume();
     }, onCancel: () {
-      if (currentSubscription != null) return currentSubscription.cancel();
```
a14n: Re-reading the `onCancel` docs I think we should keep returning. A `return null` after this line should fix the error.

lib/src/async/concat.dart
srawlins: Done.

lib/src/async/concat.dart
srawlins: Done.

srawlins:
Sorry about that; fmt'ed.

a14n:
Sorry I don't have permission to merge.

PR #478, opened 2018-09-22
Add few Iterable functions

l7ssha:
Hello, I've added few functions for `Iterable<>`.
I'm writing this kind of stuff for first time so I'm not sure that everything is fully correct.

### DIFF
--- a/lib/src/iterables/map_split.dart
+++ b/lib/src/iterables/map_split.dart
@@ -1,0 +1,93 @@
+part of quiver.iterables;
+
+typedef bool mapPred<T>(T item);
+typedef S mapFunc<T, S>(T item);
+
+/// Function executes function [func] only on element when function [pred] is true for given element.
+Iterable<T> mapWhen<T>(Iterable<T> iter, mapPred<T> pred, mapFunc<T, T> func) {
+  return iter.map((f) {
+    if (pred(f))
+      return func(f);
+    else
+      return f;
+  });
+}
+
+typedef S mapIndexFunc<T, S>(int index, T item);
+
+/// Executes func on each element
+Iterable<S> mapIndexed<T, S>(List<T> iter, mapIndexFunc<T, S> func) sync* {
+  for (var i = 0; i < iter.length; i++) {
+    yield func(i, iter[i]);
+  }
+}
+
+typedef S annotateFunc<T, S>(T item);
+
+/// Returns [Iterable] of map object which contains unmodified items of keys and result of function [func]
+Iterable<Map<T, S>> annotate<T, S>(
+    Iterable<T> iter, annotateFunc<T, S> func) sync* {
+  for (var item in iter) {
+    yield <T, S>{item: func(item)};
+  }
+}
+
+/// Return the concatenation of the result of mapping [func] over list.
+Iterable<T> mapcat<T>(Iterable<T> iter, mapFunc<T, T> func) sync* {
+  for (var item in iter) {
+    yield item;
+    yield func(item);
+  }
+}
+
+/// Returns [Iterable] which contains elements on positions from [indexes]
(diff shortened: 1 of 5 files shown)


l7ssha:
I signed it!

lib/src/iterables/map_split.dart
```diff
@@ -0,0 +1,93 @@
+part of quiver.iterables;
+
+typedef bool mapPred<T>(T item);
+typedef S mapFunc<T, S>(T item);
```
cbracken: Here and below:

Typedefs should use the same [naming convention](https://dart.dev/guides/language/effective-dart/style#do-name-types-using-uppercamelcase) as other types. Generally, [avoid using abbreviations](https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations) and favour full words. We should also use Dart 2 style typedefs (I've got a separate patch that migrates existing code to that style):
```dart
typedef MapPredicate<T> = bool Function(T item);
typedef MapFunction<T, S> = S Function(T item);
```

lib/src/iterables/map_split.dart
```diff
@@ -0,0 +1,93 @@
...
+    if (prefix[i] == iter[i]) matching++;
+  }
+
+  return matching == prefix.length;
+}
+
+/// Returns true if [Iterable] [iter] ends with elements from [prfix]
+bool isSuffix<T>(List<T> suffix, List<T> iter) {
+  return (isPrefix(suffix, iter.sublist(iter.length - suffix.length)));
+}
```
cbracken: This looks like you'll get an error when `suffix.length > iter.length`. For performance reasons, implement this in full rather than delegating to `isPrefix`.

lib/src/iterables/map_split.dart
```diff
@@ -0,0 +1,93 @@
...
+bool isPrefix<T>(List<T> prefix, List<T> iter) {
+  int matching = 0;
+  for (var i = 0; i < prefix.length; i++) {
+    if (prefix[i] == iter[i]) matching++;
+  }
+
+  return matching == prefix.length;
+}
+
+/// Returns true if [Iterable] [iter] ends with elements from [prfix]
```
cbracken: /// Returns true if [iter] ends with the elements of [suffix] or if [suffix] is empty.

lib/src/iterables/map_split.dart
```diff
@@ -0,0 +1,93 @@
...
+
+  for (var item in iter) {
+    if (func(item))
+      one.add(item);
+    else
+      two.add(item);
+  }
+
+  return [one, two];
+}
```
cbracken: Similar argument to above -- the ergonomics of this seem less useful than just:
```dart
List first = list.where(func)
List second = list.where((item) => !func(item));
```
In the above example, `first` and `second` are lazily evaluated which may or may not be an advantage depending on the code in question. If `func` is expensive and you evaluate all elements of the results, it'll have twice the cost of your eagerly-evaluated version.

cbracken:
Apologies for the delay here. As noted, most of these are unlikely to be merged. The typo is something we should absolutely merge though -- feel free to send that in a separate PR. `isPrefix` and `isSuffix` seem potentially useful if they were generalised to take Iterables rather than lists, though as I don't think the core libs have anything that's a quick one-line drop in for those.

l7ssha:
I wrote this entire thing back when these function wasn't there yet, since they are available in sdk, as you said there is no point in this. I think it's to close

yjbanov:
@l7ssha this looks like a handy collection of functions. I definitely needed something like `mapWhen` a few times. However, we're trying to minimize the feature set in this package because most of the stuff can be done with a couple of lines of code. Adding new features would take us in the opposite direction.

So I'm going to close this PR, but perhaps these functions could be added to another package, or maybe even a blog post about the various algorithms you can do with collections in Dart.

PR #479, opened 2018-09-30
Added annotations method

l7ssha:
Hello, I've added two method which allows to get access to element annotations.


yjbanov:
We do not intend to expand on mirrors functionality as that library is practically deprecated.

Issue #482, opened 2018-10-31
Add Optional.isNotPresent to quiver.core


kostaa:
Would it be possible to add `Optional.isNotPresent` getter? I find myself using the `false == opt.isPresent` far more often then `opt.isPresent`.

yjbanov:
What do you think of `!opt.isPresent`?

kostaa:
Yeah that is always possible but tired eyes sometimes miss that slim explanation mark so I prefer to use the more in your face form :-) .

Anyway my suggestion just follows similar patters - eg collections isEmpty/isNotEmpty....

zoechi:
it's like `isNotEmpty`

cbracken:
Seems like a reasonable to me! Sent #483.

PR #483, opened 2018-11-05
Adds Optional.isNotPresent

cbracken:
isNotPresent returns true when called on an Optional that lacks a value.
Analogous to isEmpty/isNotEmpty property pairs on other classes.


cbracken:
Resolves #482.

cbracken:
Fast! Thanks!

PR #484, opened 2018-11-05
Add info message when not running coverage

cbracken:
We require both a coverage token Dart dev SDK build to run coverage.
This just adds a short info message to make it clear coverage isn't run
in other cases.


yjbanov:
lgtm

PR #488, opened 2019-01-03
Adding generic types to partition

lucyget:
Changing the partition function to be generic so that types aren't lost

### DIFF
--- a/lib/src/iterables/partition.dart
+++ b/lib/src/iterables/partition.dart
@@ -45,8 +45,9 @@ class _PartitionIterator<T> implements Iterator<List<T>> {
   bool moveNext() {
     var newValue = <T>[];
     var count = 0;
-    for (; count < _size && _iterator.moveNext(); count++) {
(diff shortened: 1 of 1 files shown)


lib/src/iterables/partition.dart
```diff
@@ -15,35 +15,35 @@
...
   @override
-  List get current => _current;
+  List<T> get current => _current;
 
   @override
   bool moveNext() {
-    var newValue = [];
+    var newValue = <T>[];
     var count = 0;
     for (; count < _size && _iterator.moveNext(); count++) {
```
yjbanov: Could you please also change this `for` loop to a `while` loop while you're here? It's unreadable.

yjbanov:
Thanks for the fix!

lgtm

lib/src/iterables/partition.dart
lucyget: Done!

PR #491, opened 2019-03-26

creisman:

I've cloned this from google#474.

### DIFF
--- a/test/cache/map_cache_test.dart
+++ b/test/cache/map_cache_test.dart
@@ -67,20 +67,22 @@ main() {
     });
 
     test("should not make multiple requests for the same key", () async {
+      final completer = Completer<String>();
       int count = 0;
 
       Future<String> loader(String key) {
         count += 1;
-        return new Future.delayed(
-            const Duration(milliseconds: 1), () => "test");
+        return completer.future;
       }
 
-      await Future.wait([
+      final futures = Future.wait([
         cache.get("test", ifAbsent: loader),
         cache.get("test", ifAbsent: loader),
       ]);
 
+      completer.complete('bar');
       expect(count, equals(1));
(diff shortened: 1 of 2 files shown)


creisman:
The Travis shard, dartdevc_test is failing with the same error at HEAD (and seems to be independent of me).

test/cache/map_cache_test.dart
```diff
@@ -66,20 +67,39 @@ main() {
...
+        throw new StateError("Request failed");
+      }
+
+      await expectLater(
+          await () => cache.get("test", ifAbsent: loader), throwsStateError);
+      await expectLater(
+          await () => cache.get("test", ifAbsent: loader), throwsStateError);
+
+      expect(count, equals(2));
+      expect(await cache.get('test'), isNull);
```
yjbanov: Could you also follow-up by a third attempt that succeeds and verify that the cache successfully returns a value?

test/cache/map_cache_test.dart
```diff
@@ -15,8 +15,9 @@
 library quiver.cache.map_cache_test;
 
 import 'dart:async';
-import 'package:test/test.dart';
+
 import 'package:quiver/cache.dart';
+import 'package:test/test.dart';
```
yjbanov: nit: we put 3rd-party imports before local imports, so the previous ordering was correct.

test/cache/map_cache_test.dart
creisman: Done

test/cache/map_cache_test.dart
creisman: Done

creisman:
The remaining failing test is failing at HEAD.

creisman:
@yjbanov tagging as you said notifications weren't working right.

PR #494, opened 2019-04-03
Upgrade build dependencies

yjbanov:
I was told that this should solve the problem of failing Travis builds.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -26,9 +26,9 @@ dependencies:
 dev_dependencies:
   path: '>=1.0.0 <2.0.0'
   test: '>=1.2.0 <2.0.0'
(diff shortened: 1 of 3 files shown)


yjbanov:
This is the explanation I was given: https://github.com/dart-lang/sdk/issues/36264

However, it seems my change isn't sufficient. We seem to rely on a `pub get` supplied by the `language: dart` Travis configuration. So I'll need to inject a `pub upgrade` earlier in the process.

yjbanov:
Actually, nvm, `pub upgrade` does happen before we run the tests. The issue is that we're picking up an ancient version of `build_modules`.

yjbanov:
So, the fix was to upgrade our build dependencies. PTAL.

PR #497, opened 2019-04-25
Disable VM service port auth codes

kevmoo:
Coverage is currently broken due to recent VMs shipping with service port security auth codes enabled by default. This disables auth codes until the coverage package is updated to handle them.

Also fixes an extraneous `build` stage in travis.yml.


cbracken:
Thank you, fine sir!

PR #498, opened 2019-05-22
Narrow generate_for glob in build.yaml

jakemac53:
The existing generate_for was overriding the default glob to also include the `lib` directory, which was slowing down initial builds a tiny bit for all users who have a transitive dep on this package. (it wouldn't actually compile them since they aren't apps, but it was running on all dart files in lib and skipping them).

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,7 +1,0 @@
-targets:
-  $default:
-    builders:
-      build_web_compilers|entrypoint:
-        generate_for:
-          include: ["test/**_test.dart"]
-          exclude: ["test/all_tests.*", "test/mirrors_test.*"]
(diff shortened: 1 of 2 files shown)


yjbanov:
lgtm

jakemac53:
@yjbanov fwiw you can remove the `build.yaml` entirely also if you bump `build_web_compilers` to `^2.0.0`. I tested locally and the tests also compile twice as fast.

jakemac53:
would you all prefer this change or me updating it to remove the `build.yaml` entirely and bump the dependencies?

cbracken:
If you can remove the build.yaml completely, I'm happy :)

jakemac53:
Done!

cbracken:
re-lgtm!

yjbanov:
Re-LGTM

PR #500, opened 2019-08-01
Instrument pending timers in tests

### LINKED ISSUE #421: FakeAsync should allow you to get the stack trace of where active timers were created

jamesderlin:
Flutter widget tests assert if a test completes with timers still
pending.  However, it can be hard to diagnose where a pending timer
came from.  For example, a widget might consume a third-party library
that internally uses a timer.

To facilitate this, modify quiver's FakeAsync to store
StackTrace.current when constructing each _FakeTimer.  Add a
pendingTimersDebugInfo getter to to report this.

Fixes #421.

### DIFF
--- a/lib/testing/src/async/fake_async.dart
+++ b/lib/testing/src/async/fake_async.dart
@@ -107,6 +107,9 @@ abstract class FakeAsync {
       {Duration timeout: const Duration(hours: 1),
       bool flushPeriodicTimers: true});
 
+  /// Debugging information for all pending timers.
+  List<String> get pendingTimersDebugInfo;
+
   /// The number of created periodic timers that have not been canceled.
   int get periodicTimerCount;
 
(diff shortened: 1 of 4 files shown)


lib/testing/src/async/fake_async.dart
```diff
@@ -292,4 +301,10 @@ class _FakeTimer implements Timer {
   int get tick {
     throw new UnimplementedError("tick");
   }
+
+  /// Returns debugging information to try to identify the source of the
+  /// [Timer].
+  String get debugInfo =>
```
yjbanov: Should we preserve the `StackTrace` to allow downstream code to decide how to filter stack frames?

jamesderlin: I think preserving it might be more trouble than it's worth.  Downstream code currently can filter by discarding the first line and using `StackTrace.fromString`; I probably should have explicitly made this part of the contract.

PR #502, opened 2019-08-02
Better document the format for FakeAsync.pendingTimersDebugInfo

jamesderlin:
Additionally tweak the corresponding tests.


cbracken:
Looks like the test is failing.

jamesderlin:
> Looks like the test is failing.

Oops, apparently StackTrace does not have consistent string representations between VM and JavaScript.  I've reverted that part.

PR #508, opened 2019-09-26
Replace deprecated isInstanceOf with isA

cbracken:
In package:matcher v0.12.5, an `isA` function was added, which returns a
matcher that matches instances of the specified type and its subclasses.
This replaces the (now deprecated) `isInstanceOf` class.

This also bumps the package version to 2.0.6-dev.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: quiver
-version: 2.0.5
+version: 2.0.6-dev
 authors:
 - Justin Fagnani <justinfagnani@google.com>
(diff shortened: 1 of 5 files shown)


jtmcdole:
FakeAsync error?

yjbanov:
Look like internal DDC errors. Probably an early warning that some DDC breakage is making its way in the dev channel.

/cc @vsmenon

cbracken:
Seems I never checked this in, forgot about it, and just sent a new one. Closing :P

PR #510, opened 2019-10-01
Add `sumInts` and `sumDoubles`

mt40:
I found the sum operation is very frequently used in my projects so I want to add it here.
I tried to implement just 1 generic function like this:

```dart
T sum<T extends num>(Iterable<T> numbers) {
  ...
}
```

Howerver, I have a trouble when the input is empty. In such case, I need to return a 'zero' value. But there is no common 'zero' for both `int` and `double`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ false.
 
 `listsEqual`, `mapsEqual` and `setsEqual` check collections for equality.
 
+`sumInts` and `sumDoubles` compute sum of numbers.
+
 `LruMap` is a map that removes the least recently used item when a threshold
 length is exceeded.
 
--- a/lib/collection.dart
+++ b/lib/collection.dart
@@ -98,3 +98,13 @@ int indexOf<T>(Iterable<T> elements, bool predicate(T element)) {
   }
   return -1;
 }
+
+/// Returns the sum of the given integers.
+int sumInts(Iterable<int> numbers) {
+  return numbers.fold(0, (rs, n) => rs + n);
+}
+
+/// Returns the sum of the given doubles.
+double sumDoubles(Iterable<double> numbers) {
+  return numbers.fold(0, (rs, n) => rs + n);
+}
--- a/test/collection/collection_test.dart
+++ b/test/collection/collection_test.dart
@@ -82,4 +82,26 @@ main() {
       expect(indexOf<bool>([], (_) => true), -1);
     });
   });
+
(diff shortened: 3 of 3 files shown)


mt40:
I see and agree. Maybe this is too simple to be a util. Btw, I saw other utility libraries like this for dart and flutter around. May I ask will this repo aim to become the central repo for dart/flutter utilities?

cbracken:
Quiver was started as a 20% project by a few of us who were on the Dart team at the time and wanted a place for a bunch of core utility functionality that didn't make sense to be in the `dart:*` core libraries. Some needs we identified and landed here eventually did get implemented in the core libs -- String trim/pad methods, for example -- and we deleted them here. `List.shuffle()` did too, though I'd argue it's not likely commonly-used enough to make sense in the core libs. A decent amount of library code in Quiver was contributed from Dart codebases within Google in order to promote re-use.

As for future plans, Quiver was also designed during a time before Dart's package conventions had properly solidified, and I think in the intervening 6.5 years we've seen that the Dart ecosystem is leaning more towards smaller, more narrowly-focused packages. I suspect if it had been built today, we might have separate packages for `Multimap` or `Bimap` rather than bundling them all in Quiver, and generally I'd advise people to author new packages and let them live or die on their merits rather than add much more here, unless it's a really close fit to something we've already got.

To go on a bit of a tangent/rant:

Without speaking for others on the team, it's probably worth noting that I definitely _prefer_ Quiver/Flutter style packages over (what I'd qualify as) overly-narrowly focused packages such as what you frequently see in the web ecosystem -- I'm looking at you [left-pad](https://www.davidhaney.io/npm-left-pad-have-we-forgotten-how-to-program/). A big tradeoff being made with focused vs broad packages is the depth and complexity of the dependency tree you incur. Quiver has two, and I'll I'd admit I prefer none. One reason for narrow dependencies in the JS world is that generally, tree-shaking compilers haven't historically been broadly-used. Dart, on the other hand, doesn't have that issue: packages are a mechanism for distribution (and pub dependency), but only `import`-ed libraries incur a cost -- and even then, anything you don't use is tree-shaken away. 

My feeling is that dependency management is generally an ugly problem and that semantic versioning is, at best, an approximation of reality -- one that we've certainly screwed up versioning a few times in Quiver.

/rant

TL;DR it's not really our goal to become a central repository of Dart utilities, but we do plan to continue to maintain what's already here :)

PR #511, opened 2019-10-01
Add zip method for iterable

mt40:


cbracken:
A variant of this function already exists in the [iterables](https://github.com/google/quiver-dart/blob/master/lib/src/iterables/zip.dart#L21) library here. The existing one is slightly different in that it simply combines _n_ iterables of a given length into a single iterable of iterables, each with size _n_. If you wanted to achieve the equivalent of what this PR does, you could do:

```dart
zip([iterable1, iterable2]).map(zipper);
```

PR #513, opened 2019-10-09
Add new string methods `ifBlank()` and `ifEmpty()`

KOGI:
New string methods `ifBlank()` and `ifEmpty()` provide an easier way to set default values for empty and blank strings. These could be considered complimentary to Dart's native `??` operator (or the `??` operator could be considered as the `ifNull()` counterpart to these methods).

Example:
```dart
function doSomething(String myValue) {
  myValue ??= 'my default value'; // -- this works great for handling potential null-values
}
```

```patch
function doSomething(String myValue) {
-  myValue = (isNotEmpty(myValue) ? myValue : 'my default value');
+  myValue = ifEmpty(myValue, 'my default value'); // -- now we can treat empty/blank values in a similar way
}
```

### DIFF
--- a/lib/strings.dart
+++ b/lib/strings.dart
@@ -32,6 +32,12 @@ bool isEmpty(String s) => s == null || s.isEmpty;
 /// Returns [true] if [s] is a not empty string.
 bool isNotEmpty(String s) => s != null && s.isNotEmpty;
 
+/// Returns [s] if it's non-blank, otherwise returns [dfault]
+String ifBlank(String s, String dfault) => isBlank(s) ? dfault : s;
+
+/// Returns [s] if it's non-empty, otherwise returns [dfault]
+String ifEmpty(String s, String dfault) => isEmpty(s) ? dfault : s;
+
 /// Returns a string with characters from the given [s] in reverse order.
 ///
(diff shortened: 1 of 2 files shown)


cbracken:
Thanks for the PR! We actually debated adding something along these lines way back in 2013! [Apache commons](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html) had `defaultString`, and it seemed potentially useful, but we ended up settling on [firstNonNull](https://github.com/google/quiver-dart/blob/a4d2ef32b5ce13271d7567b66b6e158200b9cdba/lib/core.dart#L27-L33) as a more general solution to default a null string, somewhat similar to what Guava did. We should probably deprecate that function now that `??` exists.

Generally, it's highly useful (and encouraged) to eliminate nulls at the API boundaries of a given system, hence `firstNonNull` and now the `??` operator. However, after some heated discussion, we ended up moving the proposed `defaultEmpty` and `defaultBlank` functions to our 'ideas graveyard'.

On the surface, they seem potentially useful, and they certainly do reduce lines of code but in general, cases for defaulting empty (or blank) strings are significantly more nuanced -- for example, sanitising form input -- and there are strong readability arguments to making that logic explicit.

PR #518, opened 2019-10-25
Specify return in lambda passed to forEachAsync

cbracken:
forEachAsync takes a parameter of type `Future<Null>`. We now pass a
function with an explicit null return in our unit tests.


cbracken:
This cleans up analyzer hints around the existing code, but I suspect the right change is actually a breaking change to `forEachAsync` such that we expect `action` to be a void function.

Issue #242, opened 2015-03-24
Remove doWhileAsync()


justinfagnani:
It duplicates `Future.doWhile`.

The only difference is that `doWhileAsync` returns a bool indicating whether the whole iterable was processed.

cbracken:
Deprecated in #521. Will be removed in 3.0.0.

PR #521, opened 2019-10-29
Deprecate doWhileAsync

cbracken:
`doWhileAsync` is redundant with `Future.doWhile()`. We will remove this
in Quiver 3.0.0.


cbracken:
Addresses #242.

PR #522, opened 2019-11-02
Add analysis_options.yaml

cbracken:
Adds a copy of Flutter's current analysis_options.yaml with all
non-passing lints commented out with 'ENABLE'. Flutter enables a
generally well-though-out set of lints, and the authors are familiar
with this style.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,2 +1,166 @@
 analyzer:
-  strong-mode: true
+  # ENABLE strong-mode:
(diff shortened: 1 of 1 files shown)


cbracken:
This is the first of 44 PRs that get most (but not all) of the lints from Flutter enabled. Some of those marked `ENABLE` I played with and found to be of dubious value for Quiver.

cbracken:
Thanks!

PR #523, opened 2019-11-02
Add explicit return types where missing

cbracken:
Enables the always_declare_missing_return_types lint.


cbracken:
This is PR 2 of 44.

PR #524, opened 2019-11-02
Eliminate use of 'as' casts

cbracken:
Enables the avoid_as lint.

### DIFF
--- a/lib/src/async/stream_buffer.dart
+++ b/lib/src/async/stream_buffer.dart
@@ -91,7 +91,7 @@ class StreamBuffer<T> implements StreamConsumer<List<T>> {
       var subsize = leftToRead > listCap ? listCap : leftToRead;
       if (chunk is List) {
         ret.setRange(follower, follower + subsize,
-            (chunk as List<T>).getRange(_offset, _offset + subsize));
+            chunk.getRange(_offset, _offset + subsize));
       } else {
         ret[follower] = chunk;
       }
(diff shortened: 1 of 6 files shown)


cbracken:
This is PR 3 of 44.

a14n:
This lint might be deprecated in the future. See https://github.com/dart-lang/linter/issues/1401

cbracken:
Good to know. In general, I think this is still a worthwhile patch for the time being. None of these casts are required under Dart 2's type system. The one case I'm a little disappointed about is having to apply De Morgan's law to the conditional in stream_buffer.dart. It still reads fine, I guess.

lib/src/async/stream_buffer.dart
```diff
@@ -91,15 +91,15 @@ class StreamBuffer<T> implements StreamConsumer<List<T>> {
...
-            (chunk as List<T>).getRange(_offset, _offset + subsize));
+            chunk.getRange(_offset, _offset + subsize));
       } else {
         ret[follower] = chunk;
       }
       follower += subsize;
       _offset += subsize;
       _counter -= subsize;
       leftToRead -= subsize;
-      if (chunk is! List || _offset >= (chunk as List).length) {
```
yjbanov: Should we file a bug about `is!` not promoting to `List` here?

cbracken: @pq do you know if there's a bug on file for this already for the analyzer? I did a [quick search](https://github.com/dart-lang/sdk/issues?utf8=✓&q=is%3Aissue+is%3Aopen+%22is%21%22+infer+type) on "is! infer type"... guessing this must already exist though :)

pq: I'm not sure if this is being tracked anywhere.  It's possible this will get better w/ the type work for NNBD.  @stereotype441: do you happen to know?

PR #533, opened 2019-11-04
Eliminate return types on setters

cbracken:
Enables the avoid_return_types_on_setters lint.


cbracken:
9 of 44 down! It's at times like this I wish GitHub supported a diffbase mechanism where you could just export the whole stack of PRs, so you get all the benefit of independent rollback but also parallelisation.

PR #557, opened 2019-11-05
Enable the unnecessary_statements lint

cbracken:
There are a few cases in the LruMap unit tests where we want to test the
effect of accesses to an LruMap but don't care about the actual value.
We explicitly ignore this lint on those lines.

### DIFF
--- a/test/collection/lru_map_test.dart
+++ b/test/collection/lru_map_test.dart
@@ -36,7 +36,8 @@ void main() {
 
       expect(lruMap.keys.toList(), ['C', 'B', 'A']);
 
-      lruMap['B'];
+      // Trigger promotion of B.
+      final _ = lruMap['B'];
 
       // In a LRU cache, the first key is the one that will be removed if the
       // capacity is reached, so adding keys to the end is considered to be a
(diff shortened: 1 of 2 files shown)


test/collection/lru_map_test.dart
```diff
@@ -36,6 +36,7 @@ void main() {
 
       expect(lruMap.keys.toList(), ['C', 'B', 'A']);
 
+      // ignore: unnecessary_statements
       lruMap['B'];
```
a14n: You can avoid `// ignore` with:

```dart
final _ = lruMap['B'];
```

yjbanov: Either that or add a short comment just before the `// ignore` explaining why it's necessary.

cbracken: TIL :) Thanks! I've replaced the `// ignore:` with comments explaining the intent, which I admit wasn't entirely obvious to me on at least one of these. Leaving a mental note to go back and document that last one better in particular.

It's nice to be able to grep for `ignore:` and `TODO` since they're generally a code smell, so killing those off is nice. Thanks!

PR #559, opened 2019-11-05
Enable always_require_non_null_named_parameters lint

cbracken:
This requires that for any named parameter which is asserted to be
non-null is annotated with @required from package:meta.


cbracken:
Merging sans review since no code changes and just enabling a disabled lint.

PR #563, opened 2019-11-05
Use isEmpty rather than checks vs 0 where possible

cbracken:
This also adds TreeSet.isEmpty and TreeSet.isNotEmpty in order to fix a
stack overflow error when TreeSet.isEmpty or TreeSet.isNotEmpty are
called in the iterator's moveNext implementation.

Since TreeSet extends IterableBase, and didn't override isEmpty, it fell back
on the IterableBase implementation, which is to obtain an iterable over itself
and check the return value of calling moveNext. Since moveNext now calls
isEmpty, we drop into an infinite recurive loop. A quick length check
implementation is a significant performance improvement as well.

Enables the prefer_is_empty lint.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -120,7 +120,7 @@ linter:
     - prefer_inlined_adds
     # - prefer_int_literals # not yet tested
     # - prefer_interpolation_to_compose_strings # not yet tested
-    # ENABLE - prefer_is_empty
+    - prefer_is_empty
     - prefer_is_not_empty
     - prefer_iterable_whereType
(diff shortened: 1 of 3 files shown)


cbracken:
The test already exists. With the change to `isEmpty`, existing tests blow up, since all you need to do is construct an iterator. The iterator's constructor now calls isEmpty, which because it's not implemented on the collection class itself, falls back to the IterableBase implementation, which tries to construct an `Iterator` and call `moveNext`, except that `moveNext` calls `isEmpty`, causing an infinite recursive loop.

PR #568, opened 2019-11-05
Sort constructors before other members

cbracken:
While we have been relatively consistent about ordering of constructors,
enabling the sort_constructors_first lint enforces consistency.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -134,7 +134,7 @@ linter:
     - recursive_getters
     - slash_for_doc_comments
     # - sort_child_properties_last # not yet tested
-    # ENABLE - sort_constructors_first
+    - sort_constructors_first
     - sort_pub_dependencies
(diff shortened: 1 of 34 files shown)


cbracken:
@a14n @yjbanov what do you think about this? 

In some cases it seems to be a net improvement (CountdownTimer, StreamRouter are examples) and in others, it seems like a no-op. I don't see any cases I'd describe as objectively worse.

yjbanov:
I'm too used to Flutter's code style at this point. So LGTM, but I don't know if it improves readability in general :man_shrugging:

cbracken:
Yeah -- I'm also not sure if this looks better only because we've been working on Flutter for the last 4 years, or because it's a real improvement. Ah well.

a14n:
LGTM. I'm fine with this ordering.

PR #570, opened 2019-11-05
Add notes for v2.1.2 to README.md

cbracken:
Adds notes about linter-driven cleanups, as well as an improvement to
TreeSet.isEmpty/isNotEmpty performance.


cbracken:
Merged TBR since just a couple followup tweaks to readme.

PR #571, opened 2019-11-05
Version 2.1.2

cbracken:
Also adds a bit more detail in the package description.


cbracken:
Also addresses the pana complaint that our description is too short. I guess I could have just appended "Brevity is the soul of wit." to what we had.

PR #573, opened 2019-11-05
Add curly brackets for all flow control structures

cbracken:
This lint is (IMHO arbitrarily) enforced by pub.dev's Pana scoring
mechanism, so curly-brackets-on-all-the-things it is.

### DIFF
--- a/lib/src/async/concat.dart
+++ b/lib/src/async/concat.dart
@@ -67,16 +67,15 @@ class _ConcatStream<T> extends Stream<T> {
       }
     }
 
-    controller = StreamController<T>(onPause: () {
-      if (currentSubscription != null) currentSubscription.pause();
-    }, onResume: () {
-      if (currentSubscription != null) currentSubscription.resume();
-    }, onCancel: () {
-      if (currentSubscription != null)
-        return currentSubscription.cancel();
-      else
-        return null;
-    });
+    controller = StreamController<T>(
+      onPause: () {
+        currentSubscription?.pause();
+      },
+      onResume: () {
+        currentSubscription?.resume();
+      },
+      onCancel: () => currentSubscription?.cancel(),
+    );
 
     nextStream();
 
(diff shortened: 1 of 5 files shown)


lib/src/async/concat.dart
```diff
@@ -72,10 +72,11 @@ class _ConcatStream<T> extends Stream<T> {
     }, onResume: () {
       if (currentSubscription != null) currentSubscription.resume();
     }, onCancel: () {
-      if (currentSubscription != null)
+      if (currentSubscription != null) {
         return currentSubscription.cancel();
```
yjbanov: `return currentSubscription?.cancel()`

cbracken: How about this?

lib/src/async/concat.dart
```diff
@@ -67,16 +67,11 @@ class _ConcatStream<T> extends Stream<T> {
...
-      if (currentSubscription != null) currentSubscription.resume();
-    }, onCancel: () {
-      if (currentSubscription != null)
-        return currentSubscription.cancel();
-      else
-        return null;
-    });
+    controller = StreamController<T>(
+      onPause: () => currentSubscription?.pause(),
+      onResume: () => currentSubscription?.resume(),
```
yjbanov: Feels weird returning `void`, but sure.

cbracken: Hmm - the implication being from a readability pt of view that we're implying pause() and resume() have return values? Yeah, I could go either way.

yjbanov: Yeah, while all three look the same `() => expression`, only some return a non-void value.

cbracken: I admit that to my brain, which is used to years of Dart 1.0 code, this also feels naughty, but according to the Dart 2 spec, my mental model of what `=>` means needs to be updated. In section 9 on => function definitions:

> It is a compile-time error if one of the following conditions hold:
> * The function is synchronous, _T_ is not void, and it would have been a
compile-time error to declare the function with the body `{ return e; }` rather than `=> e`. _In particular, e can have any type when the return type is void. This enables concise declarations of void functions. It is reasonably easy to understand such a function, because the return type is textually near to the returned expression e._

The italicised bit is non-normative text. Anyway, after looking at this some more I think I agree. Switched the void bits back to non-fat-arrow form.

PR #587, opened 2020-02-28
Eliminate bad use of const in Optional.transform

cbracken:
It's invalid to create a const which references a type variable. The
language team decided that in order to avoid painful migration of common
Dart 1 usages such as `const []`, rather than generate an error, they
would replace the type variable with `Null`, since usually, for properly
covariant types, the result 'works'.

In cases where `transform` is chained with e.g., `or`, you can end up
with `Optional<T>.absent` replacing `T` with `Null`, but then do an
implicit covariance check on the incoming type in `or`, which should be
the `S` in `transform` but is found to be `Null`.

As a result, we revert the move to const construction in `transform` and
`transformNullable`.

### DIFF
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -89,7 +89,7 @@ class Optional<T> extends IterableBase<T> {
   /// The transformer must not return [null]. If it does, an [ArgumentError] is thrown.
   Optional<S> transform<S>(S transformer(T value)) {
     return _value == null
-        ? const Optional.absent()
+        ? Optional<S>.absent()
         : Optional.of(transformer(_value));
   }
 
@@ -100,7 +100,7 @@ class Optional<T> extends IterableBase<T> {
   /// Returns [absent()] if the transformer returns [null].
   Optional<S> transformNullable<S>(S transformer(T value)) {
     return _value == null
-        ? const Optional.absent()
+        ? Optional<S>.absent()
         : Optional.fromNullable(transformer(_value));
   }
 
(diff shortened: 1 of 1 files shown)


lib/src/core/optional.dart
```diff
@@ -89,7 +89,7 @@ class Optional<T> extends IterableBase<T> {
   /// The transformer must not return [null]. If it does, an [ArgumentError] is thrown.
   Optional<S> transform<S>(S transformer(T value)) {
     return _value == null
-        ? const Optional.absent()
+        ? Optional<S>.absent()
         : Optional.of(transformer(_value));
```
yjbanov: Maybe here too for consistency?

lib/src/core/optional.dart
```diff
@@ -89,7 +89,7 @@ class Optional<T> extends IterableBase<T> {
   /// The transformer must not return [null]. If it does, an [ArgumentError] is thrown.
   Optional<S> transform<S>(S transformer(T value)) {
     return _value == null
-        ? const Optional.absent()
+        ? Optional<S>.absent()
```
yjbanov: One way to avoid the extra allocation is to keep a static `Map<Type, Optional>` cache. Since there's an upper bound for the number of types in any given Dart program this map won't grow indefinitely. Of course, that adds a map look-up.

lib/src/core/optional.dart
cbracken: Done.

cbracken:
> this is a breaking change (breaks identity of empty optionals). this allocates more objects.

True. This reverts a previous broken change introduced in e7690454.

lib/src/core/optional.dart
cbracken: It adds a map lookup and a static cache. That said this just reverts back to the same number of allocations we did before e769045.

yjbanov:
LGTM. Didn't realize this is what we had before.

PR #593, opened 2020-04-13
Add sum for Iterable

### LINKED ISSUE #592: [Feature request] sum of Iterable
Would save some keystrokes using it instead of writing `i.reduce((a, b) => a + b)` everytime.

Anrock:
Closes #592

### DIFF
--- a/lib/iterables.dart
+++ b/lib/iterables.dart
@@ -27,3 +27,4 @@ part 'src/iterables/partition.dart';
 part 'src/iterables/generating_iterable.dart';
 part 'src/iterables/range.dart';
 part 'src/iterables/zip.dart';
+part 'src/iterables/sum.dart';
--- a/lib/src/iterables/sum.dart
+++ b/lib/src/iterables/sum.dart
@@ -1,0 +1,7 @@
+part of quiver.iterables;
+
+/// Returns the sum of values in [i] or `null` if [i] is empty.
+T sum<T extends num>(Iterable<T> i) {
+  if (i.isEmpty) return null;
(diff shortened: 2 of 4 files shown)


Anrock:
Small thing I've noticed in CONTRIBUTIND.md:
`If you're not already listed as an author in pubspec.yaml, remember to add yourself and claim your rightful place amongst the Quiverati.`. There is no `authors` section in pubspec.yaml. Is this point out of date?

cbracken:
Thanks for your contribution! I think this is probably a bit too specific to take since it applies only to types that support a `+` operator (and realistically, probably just numeric types) so it's not as general as e.g. `partition` or `zip`, or `merge`, or `concat`. The closest thing might be our `min` and `max` implementations but those take a comparator, and honestly I'm somewhat on the fence as to the merit of our `min` and `max` implementations in the first place.

This is a pretty reasonable feature to want though -- I'd be tempted to add this to your own project (or as a separate library) but implement as an [extension method](https://dart.dev/guides/language/extension-methods) like this:

```dart
extension SumIterable on Iterable<T> {
  int get sum = this.fold(0, (x, y) => x + y);
}
```

quick warning that I didn't test that at all :)

> There is no authors section in pubspec.yaml. Is this point out of date?

Oh wow! Nice catch! Thanks for spotting that. The `authors` section was deprecated by pub a while back. Authors should probably live in the AUTHORS file. If you want to patch the CONTRIBUTING.md file, I'd be thrilled to take a PR (and add yourself to the AUTHORS list as part of it!)

PR #595, opened 2020-04-13
Add avg for Iterable

### LINKED ISSUE #594: [Feature request] average of Iterable
Well, sometimes you need it.

Anrock:
Closes #594

### DIFF
--- a/lib/src/iterables/avg.dart
+++ b/lib/src/iterables/avg.dart
@@ -1,0 +1,7 @@
+part of quiver.iterables;
+
+/// Returns the average of values in [i] or `null` if [i] is empty.
+double avg<T extends num>(Iterable<T> i, [Comparator<T> compare]) {
+  if (i.isEmpty) return null;
+  return i.reduce((a, b) => a + b) / i.length;
+}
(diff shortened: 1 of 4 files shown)


lib/src/iterables/avg.dart
```diff
@@ -0,0 +1,7 @@
+part of quiver.iterables;
+
+/// Returns the average of values in [i] or `null` if [i] is empty.
+double avg<T extends num>(Iterable<T> i, [Comparator<T> compare]) {
+  if (i.isEmpty) return null;
+  return i.reduce((a, b) => a + b) / i.length;
```
Anrock, line 6: Could use `sum` from #593 if it would be merged before this

test/iterables/avg_test.dart
```diff
@@ -0,0 +1,16 @@
+library quiver.iterables.avg_test;
+
+import 'package:test/test.dart';
+import 'package:quiver/iterables.dart';
+
+void main() {
+  group('avg', () {
+    test('should return the average of elements', () {
+      expect(avg([2, 5, 1, 4]), 3);
```
Anrock, line 9: Test passes, but I'm not sure if it's a good idea to use strict equality with doubles in Dart. Is there any better/idiomatic way?

cbracken:
Thanks for your contribution! 

Like sum in #593 I think this is probably a bit too specific to take since it applies only to types that support a + and / operator (and realistically, probably just numeric types) -- see discussion on that PR. As noted there, a nice approach might be an extension method.

PR #597, opened 2020-06-29
Add removeWhere method in Multimap

Sergio-Vila:
Add in Quiver Multimap the same removeWhere method available in core:Map (https://api.dart.dev/stable/2.8.4/dart-core/Map/removeWhere.html).

### DIFF
--- a/test/collection/multimap_test.dart
+++ b/test/collection/multimap_test.dart
@@ -417,6 +417,34 @@ void main() {
       expect(map.containsKey('k2'), true);
     });
 
+    test('should call removeWhere with all {key, value} pairs', () {
+      Set s = Set();
+      ListMultimap<String, String>()
+        ..add('k1', 'v1')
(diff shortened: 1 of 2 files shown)


test/collection/multimap_test.dart
```diff
@@ -417,6 +417,34 @@ void main() {
...
+    test(
+        'should remove all the {key, value} pairs that satisfy the '
+        'predicate in removeWhere', () {
+      var map = ListMultimap<String, String>()
+        ..add('k1', 'v1')
+        ..add('k1', 'v2')
+        ..add('k2', 'v1')
+        ..removeWhere((k, v) => k == 'k2' || v == 'v1');
+      expect(map.keys, equals(['k1']));
+      expect(map.values, equals(['v2']));
```
cbracken, line 445: If you wanted to be extra-paranoid check that values for `k1` equals `['v2']`.

PR #603, opened 2020-08-24
Fix unnecessary use of null as initial value

cbracken:
When testing Metronome, we maintain a lastTime variable which we update
on each beat of the metronome, then later test against expected updates.
We were previously initializing this to null, but it's equally
reasonable to set it to the clock start time as the time at 'zero-th'
(minus-first?) beat and avoid the null.


cbracken:
This is work towards #606.

PR #604, opened 2020-08-24
NNBD: Late initialization of MapCache in tests

cbracken:
From the perspective of nullability, this avoids initialisation to null.
From the perspective of readability, initialisation now occurs closer to
use so it's clear to the reader which constructor is being called.

### DIFF
--- a/test/cache/map_cache_test.dart
+++ b/test/cache/map_cache_test.dart
@@ -20,7 +20,7 @@ import 'package:quiver/cache.dart';
 
 void main() {
   group('MapCache', () {
-    MapCache<String, String> cache;
+    /*late*/ MapCache<String, String> cache;
 
     setUp(() {
       cache = MapCache<String, String>();


test/cache/map_cache_test.dart
```diff
@@ -20,19 +20,15 @@ import 'package:quiver/cache.dart';
 
 void main() {
   group('MapCache', () {
-    MapCache<String, String> cache;
```
jtmcdole, line 23: Could this have just been:

```dart
var cache = MapCache<String, String>()
```

Less change in the file, still safe for nullability.

cbracken: Ah I see, just initialise it once here, to avoid the nullability concern, and leave the `setUp` code as-is to reset it?

jtmcdole: Yeah, I don't know what's really best. It feels about the same way of reading though, if backwards.

cbracken: Went with door #3, late initialization!

PR #605, opened 2020-08-24
NNBD: Late initialization of shared test variables

cbracken:
From the perspective of nullability, this avoids initialization to null.


cbracken:
This is work towards #606.

Issue #606, opened 2020-09-02
Add nullability annotations for non-null by default migration


cbracken:
In preparation for [NNBD](https://github.com/dart-lang/language/issues/110) in Dart, we'll need to add nullability annotations throughout Quiver.

Steps:
* [x] Apply migration tool hiniting to all libraries/tests with no cross-dependencies on other quiver libraries. This allows us to maximise our use to the `dart migrate` tool, since (as far as I can tell), it's not possible to run it over a partially migrated package.
* [x] Add language version annotation of `// @dart 2.9` to remaining files
* [x] Land migration of all hinted files; update infra to set `--enable-experiment=non-nullable`.
* [x] Migrate remaining libraries one at a time, removing language version annotations.

Libraries to migrate:
* [x] async/collect (no change necessary)
* [x] async/concat (#644)
* [x] async/countdown_timer (#656)
* [x] async/enumerate (no change necessary)
* [x] async/future_stream (#657)
* [x] async/iteration (#658)
* [x] async/metronome (#661)
* [x] async/stream_buffer (#659)
* [x] async/stream_router (#660)
* [x] async/string (no change necessary)
* [x] cache (#654)
* [x] check (#638)
* [x] collection/bimap (#650)
* [x] collection/delegates (#641)
* [x] collection/lrumap (#651)
* [x] collection/multimap (#655)
* [x] collection/treeset (#662)
* [x] collection/utils (#649)
* [x] core (#633, #652)
* [x] iterables (#643)
* [x] pattern (#639)
* [x] strings (#640)
* [x] time (no change necessary)
* [x] testing/async (#642)
* [x] testing/equality (#642)
* [x] testing/time (#642)

rokob:
I'd like to help out here if I can because I'd like core to get marked as non-nullable so that tuple can get migrated so that I can then migrate some of my libraries that depend on those two. If I understand correctly, you can still be backwards compatible with pre-NNBD SDKs even if you opt-in a file with the new SDK annotation, but you just don't get all the benefits until everyone opts-in. So why not just incrementally opt-in files rather than committing all of the comments?

cbracken:
@rokob Sorry for the delay -- I missed your comment earlier. I've updated the steps above in an effort to clarify the rationale for checking in the hinted files for as long as possible. It comes down to maximising our use of the migration tool which, as far as I can tell, doesn't support running on partially-migrated packages, nor does it support running on a single library at a time within a package.

[...]

PR #607, opened 2020-09-02
Add null annotations to core.Optional

cbracken:
This adds nullability assertions to Optional in preparation for
migration to non-null by default in Dart.

The reason that we extend IterableBase<T> rather than IterableBase<T?>
is that Optional.expand eliminates any nulls.

Once NNBD is available, the utility of this class is limited and we
should consider deprecating it.

This is work towards #606.

### DIFF
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -26,27 +26,27 @@ class Optional<T> extends IterableBase<T> {
   /// Constructs an Optional of the given [value].
   ///
   /// Throws [ArgumentError] if [value] is null.
-  Optional.of(T value) : _value = value {
+  Optional.of(T /*!*/ value) : _value = value {
     if (_value == null) throw ArgumentError('Must not be null.');
   }
 
   /// Constructs an Optional of the given [value].
   ///
   /// If [value] is null, returns [absent()].
-  const Optional.fromNullable(T value) : _value = value;
+  const Optional.fromNullable(T /*?*/ value) : _value = value;
 
-  final T _value;
+  final T /*?*/ _value;
 
   /// True when this optional contains a value.
-  bool get isPresent => _value != null;
+  bool /*!*/ get isPresent => _value != null;
 
   /// True when this optional contains no value.
-  bool get isNotPresent => _value == null;
+  bool /*!*/ get isNotPresent => _value == null;
 
   /// Gets the Optional value.
   ///
   /// Throws [StateError] if [value] is null.
-  T get value {
+  T /*!*/ get value {
     if (_value == null) {
       throw StateError('value called on absent Optional.');
     }
@@ -54,7 +54,7 @@ class Optional<T> extends IterableBase<T> {
   }
 
   /// Executes a function if the Optional value is present.
-  void ifPresent(void ifPresent(T value)) {
+  void ifPresent(void ifPresent(T /*!*/ value)) {
     if (isPresent) {
       ifPresent(_value);
     }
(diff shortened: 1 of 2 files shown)


cbracken:
@davidmorgan for review too since he's the author of this code.

lib/src/core/optional.dart
```diff
@@ -46,15 +46,15 @@ class Optional<T> extends IterableBase<T> {
...
+  T/*!*/ get value {
     if (_value == null) {
       throw StateError('value called on absent Optional.');
     }
     return _value;
   }
 
   /// Executes a function if the Optional value is present.
-  void ifPresent(void ifPresent(T value)) {
+  void ifPresent(void ifPresent(T/*!*/ value)) {
```
cbracken, line 57: Here and the transform methods below -- we should annotate the function params as `/*!*/` but I need to look up how :)

a14n: AFAICT you cannot use `T!`. You have to make T non nullable with `Optional<T extends Object>` and use `T?` where a nullable value can appear.

cbracken: Interesting. Alright will do some experimenting with this. Thanks @a14n.

yjbanov (approved):
LGTM if LGT @a14n

a14n (approved):

lib/src/core/optional.dart
```diff
@@ -26,35 +26,35 @@ class Optional<T> extends IterableBase<T> {
   /// Constructs an Optional of the given [value].
   ///
   /// Throws [ArgumentError] if [value] is null.
-  Optional.of(T value) : _value = value {
+  Optional.of(T /*!*/ value) : _value = value {
     if (_value == null) throw ArgumentError('Must not be null.');
```
davidmorgan, line 30: Null checks on T! aren't allowed in null safe mode, because they're not necessary. But, they _are_ needed in mixed mode. The recommended way is to use

https://api.flutter.dev/flutter/dart-core/ArgumentError/checkNotNull.html

lib/src/core/optional.dart
```diff
@@ -72,22 +72,22 @@ class Optional<T> extends IterableBase<T> {
   /// The default is returned if the Optional is [absent()].
   ///
   /// Throws [ArgumentError] if [defaultValue] is null.
-  T or(T defaultValue) {
+  T /*!*/ or(T /*!*/ defaultValue) {
     if (defaultValue == null) {
       throw ArgumentError('defaultValue must not be null.');
```
davidmorgan, line 77: ArgumentError.checkNotNull

lib/src/core/optional.dart
```diff
@@ -98,26 +98,27 @@ class Optional<T> extends IterableBase<T> {
...
 
   /// Delegates to the underlying [value] hashCode.
   @override
-  int get hashCode => _value.hashCode;
+  int /*!*/ get hashCode => _value.hashCode;
 
   /// Delegates to the underlying [value] operator==.
   @override
-  bool operator ==(o) => o is Optional<T> && o._value == _value;
+  bool /*!*/ operator ==(o) => o is Optional<T> && o._value == _value;
```
davidmorgan, line 118: `operator==` now takes `Object o` to indicate that it is never passed null.

davidmorgan (approved):
"T extends Object" sounds good.

FWIW, so far I've been favouring actually migrating file by file (with version comments to hold back non-migrated code) rather than checking in comments. The main benefits this gets you are the full null safe analyzer/static checks and actual test runs. At the same time you might want to start running tests both with and without null safety and/or null safe asserts enabled.

lib/src/core/optional.dart
yjbanov: While `T!` is not allowed, `T/*!*/` is used by the migration tool as a signal that means "do not make this nullable", and is converted to just `T`.

cbracken:
Replaced by #633.

PR #611, opened 2020-10-23
Add Directory type check in visitDirectory

cbracken:
In visitDirectory in the IO library, we assume that if a filesystem
entity is not a File or a Link, that it's a Directory. This adds a type
check to ensure that's actually the case.


jtmcdole:
lgtm

PR #614, opened 2020-10-24
Simplify bailout in visitDirectory

cbracken:
Simplify the null check in visitDirectory and add a TODO to make it
easier to remove post NNBD migration.


jtmcdole:
Should we also make this function `async / await` at some point?

jtmcdole:
lgtm

PR #616, opened 2020-10-24
Use ArgumentError.checkNotNull where possible

cbracken:
This extracts null checks that we'll be able to remove after non-null by default migration.

This bumps the minimum SDK constraint to 2.10 and deprecates checkNotNull and firstNonNull.

Note: this patch is against the (new) `null_safety` branch.

Bug: https://github.com/google/quiver-dart/issues/606

### DIFF
--- a/lib/strings.dart
+++ b/lib/strings.dart
@@ -68,8 +68,10 @@ String _reverse(String s) {
 /// loop('test.txt', -3) == 'txt'
 /// loop('ldwor', -3, 2) == 'world'
 String loop(String s, int from, [int to]) {
-  if (s == null || s == '') {
-    throw ArgumentError('Input string cannot be null or empty');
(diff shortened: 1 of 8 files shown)


lib/strings.dart
```diff
@@ -68,8 +68,10 @@ String _reverse(String s) {
 /// loop('test.txt', -3) == 'txt'
 /// loop('ldwor', -3, 2) == 'world'
 String loop(String s, int from, [int to]) {
-  if (s == null || s == '') {
-    throw ArgumentError('Input string cannot be null or empty');
+  // TODO(cbracken): Remove post-NNBD. https://github.com/google/quiver-dart/issues/612
+  ArgumentError.checkNotNull(s);
+  if (s == '') {
```
a14n, line 73: `s.isEmpty`

a14n (approved):

lib/strings.dart
cbracken: Done. Thanks!

PR #624, opened 2020-10-26
Switch from parts to re-exports of libraries

cbracken:
Migrates from using part/part of directives to re-exporting the private
libraries from the top-level lib/async.dart.


yjbanov (approved):
lgtm

Issue #621, opened 2020-10-24
Consider deprecating and removing IO/mirrors libraries


cbracken:
The IO library contains three functions. I did a quick scan of internal usage at Google and found that across hundreds of thousands of imports of quiver, these have almost no usage whatsoever; probably because most of our code is either web or Flutter.

Eliminating this has the nice effect of making Quiver show as a web-friendly package on pub.dev. If we also eliminated or split out the mirrors package it would show as flutter-compatible too. If we think they're worth keeping perhaps we could split them out into a separate package, or find them a home in an existing package.

## byteStreamToString
```dart
//  Converts a [Stream] of byte lists to a [String].
Future<String> byteStreamToString(Stream<List<int>> stream, {Encoding encoding = utf8}) {
  return stream.transform(encoding.decoder).join();
}
```
I found one usage in Fuchsia.

This doesn't actually depend on dart:io, but I suspect we hadn't yet created the async library when this was landed. If we want to keep this, we can move it to the async library, forward it here and deprecate this one until we delete it.

##  getFullPath
```dart
String getFullPath(path) => File(path).resolveSymbolicLinksSync();
```
I found no uses of this internally.

This is potentially useful in a case like `listOfPaths.map(getFullPath)` but it's also a one-liner that existing users could easily inline into their code if they like it.

## visitDirectory

```dart
/// Lists the sub-directories and files of this Directory, optionally recursing
/// into sub-directories based on the return value of [visit].
///
/// [visit] is called with a [File], [Directory] or [Link] to a directory,
/// never a Symlink to a File. If [visit] returns true, then its argument is
/// listed recursively.
Future visitDirectory(Directory dir, Future<bool> visit(FileSystemEntity f)) {
  ...
}
```
I found no uses of this internally.

This is potentially useful, but the fact that we don't use it, even within our Dart tooling, suggests that it's probably not worth keeping around. If there's another package where it might be a good fit we could contribute it there.

/cc @yjbanov @jtmcdole @adambender @a14n for thoughts.

jtmcdole:
I've written a "visitDirectory" I think about twice in the past, and I'd
probably just copy-paste that anytime I needed it. If we had a
package:filesystem, it would go great there... not generally a quivery
thing though. Also, the "visitor" pattern is well known.

`collectStream<T>` could just go in async. The number of times I've wanted
to go from Stream<List<T>> to List<T>....

cbracken:
Closed by #628 and #629.

[...]

PR #628, opened 2020-10-26
Move stringFromByteStream to async; deprecate IO

cbracken:
This deprecates two of the three function in the IO library and migrates
the remaining function (which doesn't use dart:io) to the async library.

As discussed in https://github.com/google/quiver-dart/issues/621, a
survey of ~150,000 imports of quiver at Google suggests that only
stringFromByteStream is used internally. It has been moved to the async
library.

getFullPath can be replaced with File.resolveSymbolicLinksSync if used.

visitDirectory should be rewritten if any users are making use of it.
Alternatively, it and its tests can be copied to other codebases under
the terms of the Apache 2.0 license.

Eliminating these functions avoids pub.dev suggesting that quiver can't
be used for web projects.

### DIFF
--- a/lib/src/async/string.dart
+++ b/lib/src/async/string.dart
@@ -12,15 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-library quiver.async;
+import 'dart:convert' show Encoding, utf8;
 
-export 'src/async/collect.dart';
-export 'src/async/concat.dart';
-export 'src/async/countdown_timer.dart';
-export 'src/async/enumerate.dart';
-export 'src/async/future_stream.dart';
-export 'src/async/iteration.dart';
-export 'src/async/metronome.dart';
-export 'src/async/stream_buffer.dart';
-export 'src/async/stream_router.dart';
-export 'src/async/string.dart';
+///  Converts a [Stream] of byte lists to a [String].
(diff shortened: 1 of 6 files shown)


lib/src/async/string.dart
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert' show Encoding, utf8;
+
+///  Converts a [Stream] of byte lists to a [String].
+Future<String> byteStreamToString(Stream<List<int>> stream,
```
yjbanov, line 18: Do we even need it, given how short the implementation is? Maybe we should just deprecate and remove in 3.0.0?

yjbanov (approved):
LGTM, but also fine with deprecating then removing `stringFromByteStream` altogether.

lib/src/async/string.dart
cbracken: Good question -- it's used in Fuchsia in one location. It's also roughly zero-maintenance to keep around in async though.

PR #629, opened 2020-10-26
Deprecate mirrors library

cbracken:
In a survey of ~150,000 imports of quiver within Google, no uses of this           
library was found. This library was written prior to Dart 1.0 and in the           
interim, mirrors have been found to be problematic for production code.            
                                                                                   
The primary Dart runtimes today are Flutter and the web. Flutter                   
disables dart:mirrors altogether. On the web, the use of mirrors                   
disables tree-shaking since in theory almost any symbol could be used              
reflectively; this can't be solved through static analysis -- one could            
imagine the situation where a user of a web app inputs the name of a               
method to be reflectively invoked at runtime.                                      
                                                                                   
Users of this code can rewrite this code, or copy it into their own                
projects under the terms of the Apache 2.0 license.

See: https://github.com/google/quiver-dart/issues/621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@
     `File(path).resolveSymbolicLinksSync`. This will be removed in 3.0.0.
   * Deprecate `visitDirectory`. This will be removed in 3.0.0. The source can be
     copied under the terms of the Apache 2.0 license.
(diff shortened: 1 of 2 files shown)


cbracken:
This is rebased on top of #628. For review purposes, just take a look at the second commit, touching lib/mirrors.dart.

yjbanov (approved):
lgtm

PR #632, opened 2020-10-26
Enable strict linting during analysis

cbracken:
Also eliminates deprecation warnings for checkNotNull and firstNonNull
in tests.


yjbanov (approved):
lgtm

PR #633, opened 2020-10-28
Add nullability hints to core library

cbracken:
This applies nullability hints for NNBD migration to the core library.


yjbanov (approved):
lgtm

Issue #634, opened 2020-10-28
Re-enable dartfmt check when NNBD hints have been removed


cbracken:
Dartfmt really wants to format `dart migrate` nullability hint comments. Disable the format check until we apply the migration hints.

cbracken:
This was done in #664.

PR #635, opened 2020-10-28
Disable dartfmt checks during NNBD migration

cbracken:
Dartfmt really wants to format dart migrate nullability hint comments.
Disable the format check until we apply the migration hints.

Bug: https://github.com/google/quiver-dart/issues/634


yjbanov (approved):
lgtm

PR #636, opened 2020-10-28
Make firstNonNull generic

cbracken:
This allows us to declare the return value as non-null.


cbracken:
Going to close this one actually. Given that it's deprecated and we're about to kill it off, there's no sense in creating a breaking change here.

PR #637, opened 2020-10-28
Ignore vim swap files

cbracken:


cbracken:
I guess I had this in my global gitignore for years; otherwise shocked we never had this :)

yjbanov (approved):
lgtm

PR #638, opened 2020-10-28
Add nullability hints to check library

cbracken:
Applies nullability hits for NNBD migration to the check library.

Partial fix for #606.


yjbanov (approved):
lgtm

PR #639, opened 2020-10-28
Add nullability hints to pattern library

cbracken:
This applies nullability hints for NNBD migration to the pattern
library.

Partial fix for #606.

### DIFF
--- a/lib/pattern.dart
+++ b/lib/pattern.dart
@@ -31,14 +31,14 @@ String escapeRegex(String str) => str.splitMapJoin(_specialChars,
 /// Returns a [Pattern] that matches against every pattern in [include] and
 /// returns all the matches. If the input string matches against any pattern in
 /// [exclude] no matches are returned.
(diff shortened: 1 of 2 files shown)


lib/pattern.dart
```diff
@@ -109,7 +113,7 @@ class Glob implements Pattern {
   int get hashCode => pattern.hashCode;
 
   @override
-  bool operator ==(other) => other is Glob && pattern == other.pattern;
+  bool operator ==(Object other) => other is Glob && pattern == other.pattern;
```
yjbanov, line 116: Should this be `Object?`?

yjbanov (approved):
lgtm

lib/pattern.dart
cbracken: Nope-- in Dart (even as far back as 1.0) the spec says null checking is done by the runtime, so it'll never be null.

PR #640, opened 2020-10-29
Add nullability hints to strings library

cbracken:
This applies nullability hints for NNBD migration to the strings
library.

Partial fix to #606.


yjbanov (approved):
lgtm

PR #641, opened 2020-10-29
Add nullability hints to collection/delegates

cbracken:
This applies nullability hints for NNBD migration to the delegates
libraries under the broader collection library.


yjbanov (approved):
lgtm

PR #644, opened 2020-10-30
Add Nullability hints to async/concat

cbracken:
Applies nullability hits for NNBD migration to async/concat.

### DIFF
--- a/lib/src/async/concat.dart
+++ b/lib/src/async/concat.dart
@@ -40,27 +40,26 @@ class _ConcatStream<T> extends Stream<T> {
   final Iterable<Stream<T>> _streams;
 
   @override
-  StreamSubscription<T> listen(void onData(T data),
-      {Function onError, void onDone(), bool cancelOnError}) {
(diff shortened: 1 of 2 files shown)


lib/src/async/concat.dart
```diff
@@ -40,27 +40,26 @@ class _ConcatStream<T> extends Stream<T> {
...
-    StreamSubscription<T> currentSubscription;
-    StreamController<T> controller;
+    StreamSubscription<T>/*?*/ currentSubscription;
+    StreamController<T>/*?*/ controller;
     final iterator = _streams.iterator;
 
-    void nextStream() {
-      bool hasNext;
+    void nextStream(StreamController<T> controller) {
+      /*late*/ bool hasNext;
```
yjbanov, line 51: dunno if temp annotations support it, but this could be `late final`

yjbanov (approved):
lgtm

lib/src/async/concat.dart
cbracken: Will give it a shot in the upcoming patch that uncomments this!

PR #645, opened 2020-10-30
Import only what's necessary in tests

cbracken:
During migrations that affect the type system, piecewise migration is
currently made difficult by importing umbrella imports such as
quiver/async.dart. Instead, tests now import only the exact
sub-libraries they depend on.

Helps with #606.


yjbanov (approved):
lgtm

PR #646, opened 2020-10-30
Add @dart=2.9 tags to not-yet-migrated files

cbracken:
Applies `// @dart = 2.9` comments to any files that have not yet had
non-null by default migration hints added.

Progress on #606.


yjbanov (approved):
lgtm

PR #648, opened 2020-10-30
Extract utils libraries for core, collection

cbracken:
This avoids mixing implementation into umbrella libraries.


yjbanov (approved):
lgtm

PR #650, opened 2020-10-30
Migrate collection/bimap to non-null by default

cbracken:
Partial fix for #606.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
   * BREAKING CHANGE: Remove mirrors library. This was deprecated in 2.1.4.
   * BREAKING CHANGE: forEachAsync now returns Future<void> instead of
     Future<Null>.
+  * BREAKING CHANGE: BiMap no longer throws ArgumentError on insertion
+    of a null key or value, if the corresponding K or V type is
+    nullable. As before, values in the map must be unique and
+    ArgumentError is thrown on attempts to add a key-value pair whose
+    value is already in the map.
   * Deprecate `checkNotNull`. Users of this function should migrate to
     `ArgumentError.checkNotNull`. This will be removed in 4.0.0.
   * Deprecate `firstNonNull`. Users of this function should migrate to
--- a/lib/src/collection/bimap.dart
+++ b/lib/src/collection/bimap.dart
@@ -12,16 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:collection';
 
 /// A bi-directional map whose key-value pairs form a one-to-one
 /// correspondence.  BiMaps support an `inverse` property which gives access to
 /// an inverted view of the map, such that there is a mapping (v, k) for each
 /// pair (k, v) in the original map. Since a one-to-one key-value invariant
-/// applies, it is an error to insert duplicate values into this map. It is
-/// also an error to insert null keys or values into this map.
+/// applies, it is an error to insert duplicate values into this map.
 abstract class BiMap<K, V> implements Map<K, V> {
   /// Creates a BiMap instance with the default implementation.
   factory BiMap() => HashBiMap();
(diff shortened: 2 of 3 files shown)


lib/src/collection/bimap.dart
```diff
@@ -119,14 +117,16 @@ class HashBiMap<K, V> implements BiMap<K, V> {
 
   @override
   V putIfAbsent(K key, V ifAbsent()) {
-    var value = _map[key];
-    if (value != null) return value;
-    if (!_map.containsKey(key)) return _add(key, ifAbsent(), false);
-    return null;
+    if (_map.containsKey(key)) {
+      return _map[key]!;
+    }
```
yjbanov, line 122: Perhaps it doesn't matter, but the previous version was faster because in the case when the key is present we did one map lookup. The new version performs two lookups (`containsKey` then `[key]`).

Is the change because the nullness of the value is not equivalent to missing key any more?

cbracken: The previous version was correct, but the return null at the end was dead code except with Bimap<T?> which doesn't make sense.

cbracken: FWIW, the original implementation works with a little tweaking. Maybe adding a comment fixes the readability concern.
```dart
@override                                                                     
V putIfAbsent(K key, V ifAbsent()) {                                          
  V? value = _map[key];                                                       
  if (value != null) return value;                                            
  if (!_map.containsKey(key)) return _add(key, ifAbsent(), false);            
  // Value is null, which is disallowed.                                                                     
  return value!;                                                              
}                                                                             
```

cbracken: Re-pushed with that implementation.

cbracken: And re-pushed with an even better implementation. Nullable types are now disallowed for K and V so we know if V is null that K is not a key.

cbracken: Added a second commit as a "what do you think?"

There's really no reason we shouldn't allow nullable keys or values, subject to the same constraint that we've always had, that each value in the map be unique (since it's used as a key in the inverse), we'd simply be allowing one more legal value -- a single null.

test/collection/bimap_test.dart
```diff
@@ -40,6 +40,24 @@ void main() {
       expect(map.inverse.isNotEmpty, false);
     });
 
+    test('should support null keys', () {
+      final map = BiMap<String?, int>();
+      map[null] = 5;
+      expect(map.isEmpty, false);
+      expect(map.containsKey(null), true);
+      expect(map.containsValue(5), true);
+      expect(map.keys, contains(null));
```
yjbanov, line 49: Also `expect(map[null], 5);` :)

yjbanov: Might also be worth testing the contents of `map.inverse`

test/collection/bimap_test.dart
```diff
@@ -40,6 +40,24 @@ void main() {
...
+    });
+
+    test('should support null values', () {
+      final map = BiMap<String, int?>();
+      map[k1] = null;
+      expect(map.isEmpty, false);
+      expect(map.containsKey(k1), true);
+      expect(map.containsValue(null), true);
+      expect(map[k1], isNull);
```
yjbanov, line 58: Test `map.inverse`?

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
     Existing callers should migrate to `Future.doWhile()`.
   * BREAKING CHANGE: Remove IO library. This was deprecated in 2.1.4.
   * BREAKING CHANGE: Remove mirrors library. This was deprecated in 2.1.4.
+  * BREAKING CHANGE: BiMap no longer throws ArgumentError on insertion
+    of a null key or value, if the corresponding K or V type is
+    nullable. As before, values in the map must be unique and
+    ArgumentError is thrown on attempts to add a key-value pair whose
+    value is already in the map.
```
yjbanov, line 15: Do we need to update the dartdocs about null?

test/collection/bimap_test.dart
cbracken: That should be pretty heavily tested already; or a sepcific map inverse test?

CHANGELOG.md
cbracken: I deleted the bits about null from the docs.

test/collection/bimap_test.dart
yjbanov: My thinking was if you start with a bimap that allows nullable values, when you invert it you'll end up with one allowing nullable keys, which is a new thing. But perhaps it's already covered and I just can't tell if it is.

yjbanov (approved):
lgtm

test/collection/bimap_test.dart
cbracken: Nullable keys were always allowed in Dart, though like all keys only one of them. In this case the existing documentation already covers the behaviour, it's just that we allow one additional key (or value) than we previously did -- i.e., null.

test/collection/bimap_test.dart
cbracken: Done.

PR #651, opened 2020-10-30
Migrate collection/lru_map to non-null by default

cbracken:
Partial fix for #606.

### DIFF
--- a/lib/src/collection/lru_map.dart
+++ b/lib/src/collection/lru_map.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:collection';
 
 import 'package:quiver/iterables.dart' show GeneratingIterable;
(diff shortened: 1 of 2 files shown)


lib/src/collection/lru_map.dart
```diff
@@ -136,7 +135,8 @@ class LinkedLruHashMap<K, V> implements LruMap<K, V> {
 
   /// Creates an [Iterable] around the entries of the map.
   Iterable<_LinkedEntry<K, V>> _iterable() {
-    return GeneratingIterable<_LinkedEntry<K, V>>(() => _head, (n) => n.next);
+    if (_head == null) return const Iterable.empty();
```
yjbanov, line 138: Let's use `{ }`?

lib/src/collection/lru_map.dart
```diff
@@ -164,7 +164,8 @@ class LinkedLruHashMap<K, V> implements LruMap<K, V> {
 
   @override
   set maximumSize(int maximumSize) {
-    if (maximumSize == null) throw ArgumentError.notNull('maximumSize');
+    // TODO(cbracken): Remove when mixed-mode execution is unsupported.
+    ArgumentError.checkNotNull(maximumSize);
```
yjbanov, line 168: `ArgumentError.checkNotNull(maximumSize, 'maximumSize')`

lib/src/collection/lru_map.dart
```diff
@@ -284,10 +285,10 @@ class LinkedLruHashMap<K, V> implements LruMap<K, V> {
   }
 
   @override
-  V update(K key, V update(V value), {V ifAbsent()}) {
+  V update(K key, V update(V value), {V ifAbsent()?}) {
     V newValue;
     if (containsKey(key)) {
-      newValue = update(this[key]);
+      newValue = update(this[key]!);
```
yjbanov, line 291: Looks like non-nullness of `this[key]` implies `containsKey(key) == true`, so we could avoid the `!` like this:

```dart
V newValue;
final V? currentValue = this[key];
if (currentValue != null) {
  newValue = update(currentValue);
} else {
  ...
```

yjbanov (approved):
lgtm

lib/src/collection/lru_map.dart
cbracken: Done.

lib/src/collection/lru_map.dart
cbracken: Now that you need to declare nullability of the key type explicitly, we should probably allow null keys, since any use of them is much more likely to be intentional. I should probably make a similar change here as I did with BiMap and support this. Will probably do a separate PR for that though so we can discuss independently of NNBD migration.

lib/src/collection/lru_map.dart
cbracken: Done.

PR #652, opened 2020-10-31
Disallow nullable T in Optional<T>

cbracken:
The entire point of Optional is to write null-safe code. This disallows
creation of Optionals of nullable types.

Partial fix for #606.


yjbanov (approved):
lgtm

PR #654, opened 2020-10-31
Migrate cache library to non-null by default

cbracken:
This is stacked on top of https://github.com/google/quiver-dart/pull/651. Only the second commit should be reviewed.

Partial fix for #606.

### DIFF
--- a/lib/src/collection/lru_map.dart
+++ b/lib/src/collection/lru_map.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:collection';
 
 import 'package:quiver/iterables.dart' show GeneratingIterable;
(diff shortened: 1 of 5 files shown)


lib/src/collection/lru_map.dart
```diff
@@ -136,7 +135,8 @@ class LinkedLruHashMap<K, V> implements LruMap<K, V> {
 
   /// Creates an [Iterable] around the entries of the map.
   Iterable<_LinkedEntry<K, V>> _iterable() {
-    return GeneratingIterable<_LinkedEntry<K, V>>(() => _head, (n) => n.next);
+    if (_head == null) return const Iterable.empty();
```
yjbanov, line 138: `{ }`?

yjbanov (approved):
lgtm

lib/src/collection/lru_map.dart
cbracken: Done in the other PR.

PR #655, opened 2020-10-31
Migrate collection/multimap to non-null by default

cbracken:
Partial fix towards #606.

### DIFF
--- a/lib/src/collection/multimap.dart
+++ b/lib/src/collection/multimap.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:math' show Random;
 
 /// An associative container that maps a key to multiple values.
@@ -32,20 +30,20 @@ abstract class Multimap<K, V> {
       {K key(element), V value(element)}) = ListMultimap<K, V>.fromIterable;
 
   /// Returns whether this multimap contains the given [value].
-  bool containsValue(Object value);
+  bool containsValue(Object? value);
 
   /// Returns whether this multimap contains the given [key].
-  bool containsKey(Object key);
+  bool containsKey(Object? key);
 
   /// Returns whether this multimap contains the given association between [key]
   /// and [value].
-  bool contains(Object key, Object value);
+  bool contains(Object? key, Object? value);
 
   /// Returns the values for the given [key]. An empty iterable is returned if
   /// [key] is not mapped. The returned collection is a view on the multimap.
   /// Updates to the collection modify the multimap and likewise, modifications
   /// to the multimap are reflected in the returned collection.
-  Iterable<V> operator [](Object key);
+  Iterable<V> operator [](Object? key);
 
   /// Adds an association from the given key to the given value.
   void add(K key, V value);
@@ -62,11 +60,11 @@ abstract class Multimap<K, V> {
 
   /// Removes the association between the given [key] and [value]. Returns
   /// `true` if the association existed, `false` otherwise.
-  bool remove(Object key, V value);
+  bool remove(Object? key, V? value);
 
   /// Removes the association for the given [key]. Returns the collection of
   /// removed values, or an empty iterable if [key] was unmapped.
-  Iterable<V> removeAll(Object key);
+  Iterable<V> removeAll(Object? key);
 
   /// Removes all entries of this multimap that satisfy the given [predicate].
   void removeWhere(bool predicate(K key, V value));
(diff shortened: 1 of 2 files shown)


lib/src/collection/multimap.dart
```diff
@@ -798,28 +798,28 @@ class _WrappedList<K, V> extends _WrappedIterable<K, V, List<V>>
...
-  List<V> sublist(int start, [int end]) {
+  List<V> sublist(int start, [int? end]) {
     _syncDelegate();
     return _delegate.sublist(start, end);
   }
 }
 
 class _WrappedSet<K, V> extends _WrappedIterable<K, V, Set<V>>
     implements Set<V> {
-  _WrappedSet(Map<K, Iterable<V>> map, K key, Iterable<V> delegate)
```
cbracken, line 821: Funny this was never fixed but it was for _WrappedList.

lib/src/collection/multimap.dart
```diff
@@ -798,28 +798,28 @@ class _WrappedList<K, V> extends _WrappedIterable<K, V, List<V>>
...
-      : super(map, key, delegate);
+  _WrappedSet(Map<K, Iterable<V>> map, K key, Set<V> delegate)
+      : super(map as Map<K, Set<V>>, key, delegate);
```
yjbanov, line 822: Is this cast necessary? `SetMultimap<K, V> extends _BaseMultimap<K, V, Set<V>>`, which makes its `_map` field a `Map<K, Set<V>>`.

cbracken: Yep, since we're doing a narrowing cast.

yjbanov: But why does the constructor take `Map<K, Iterable<V>> map` instead of `Map<K, Set<V>> map` in the first place?

cbracken:
```dart
class _WrappedSet<K, V> extends _WrappedIterable<K, V, Set<V>>                  
    implements Set<V> {                                                         
  _WrappedSet(Map<K, Iterable<V>> map, K key, Set<V> delegate)                                                 
      : super(map, key, delegate);                                              
```

cbracken: Basically so we can avoid re-implementing everying twice.

cbracken: Discussed offline; amended the ctor generic collection type to use List<V>/Set<V> rather than Iterable<V> for C and we're all set.

lib/src/collection/multimap.dart
```diff
@@ -382,8 +381,9 @@ class _WrappedMap<K, V, C extends Iterable<V>> implements Map<K, C> {
   @override
   void removeWhere(bool test(K key, C value)) {
     var keysToRemove = <K>[];
+    // TODO(cbracken): iterate over entries.
     for (final key in keys) {
-      if (test(key, this[key])) keysToRemove.add(key);
+      if (test(key, this[key]!)) keysToRemove.add(key);
```
davidmorgan, line 386: You only want a null check if `C` is not nullable; so

`this[key]!` --> `this[key] as C`

davidmorgan (approved):
LG, thanks! A few comments.

lib/src/collection/multimap.dart
```diff
@@ -313,7 +312,7 @@ class _WrappedMap<K, V, C extends Iterable<V>> implements Map<K, C> {
   final _BaseMultimap<K, V, C> _multimap;
 
   @override
-  C operator [](Object key) => _multimap[key];
+  C? operator [](Object? key) => _multimap[key] as C;
```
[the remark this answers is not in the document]
cbracken: It's always non-null.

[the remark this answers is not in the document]
cbracken: Maybe the best way is to change it to `C?` but add a comment or an assert to document the expected behaviour for maintainers.

lib/src/collection/multimap.dart
```diff
@@ -347,9 +346,9 @@ class _WrappedMap<K, V, C extends Iterable<V>> implements Map<K, C> {
   @override
   int get length => _multimap.length;
   @override
-  C remove(Object key) => _multimap.removeAll(key);
+  C? remove(Object? key) => _multimap.removeAll(key) as C;
```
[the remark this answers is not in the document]
cbracken: I've added comments that document that _BaseMultiMap's API contract guarantees these two methods always return non-null.

yjbanov (approved):
lgtm

lib/src/collection/multimap.dart
cbracken: Good call -- thanks! Done.

PR #657, opened 2020-11-01
Migrate async/future_stream to non-null by default

cbracken:
Partial fix for #606.

### DIFF
--- a/lib/src/async/future_stream.dart
+++ b/lib/src/async/future_stream.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:async';
 
 /// A Stream that will emit the same values as the stream returned by [future]
@@ -39,8 +37,8 @@ class FutureStream<T> extends Stream<T> {
       // Since [controller] is synchronous, it's likely that emitting an error
       // will cause it to be cancelled before we call close.
       if (_controller != null) {
-        _controller.addError(e, stackTrace);
-        _controller.close();
+        _controller?.addError(e, stackTrace);
+        _controller?.close();
       }
       _controller = null;
     });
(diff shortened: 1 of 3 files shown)


lib/src/async/future_stream.dart
```diff
@@ -56,31 +54,31 @@ class FutureStream<T> extends Stream<T> {
...
-      _subscription = stream.listen(_controller.add,
-          onError: _controller.addError, onDone: _controller.close);
+      _subscription = stream.listen(_controller!.add,
+          onError: _controller!.addError, onDone: _controller!.close);
     });
   }
 
   void _onCancel() {
-    if (_subscription != null) _subscription.cancel();
+    if (_subscription != null) _subscription?.cancel();
```
jtmcdole, line 70: Why even do the null test when you can just make this:

`_subscription?.cancel();`

cbracken: I'll go hang my head in shame now. I bet I did that in stream_router too.

cbracken: Done.

jtmcdole: <3

lib/src/async/future_stream.dart
```diff
@@ -39,8 +37,8 @@ class FutureStream<T> extends Stream<T> {
       // Since [controller] is synchronous, it's likely that emitting an error
       // will cause it to be cancelled before we call close.
       if (_controller != null) {
-        _controller.addError(e, stackTrace);
-        _controller.close();
+        _controller?.addError(e, stackTrace);
```
jtmcdole, line 40: also doesn't need the surrounding check; but then again, there are two?  Would this be the case of `_controller!.addError()..close()`?

jtmcdole (approved):

lib/src/async/future_stream.dart
cbracken: You could do `_controller!..addError()..close();` but dartfmt will reformat to a three-liner. Given these have null-checks though I can kill off the conditional around this though.

cbracken: Actually I'm not sure there's a one-liner that's possible regardless of dartfmt. You could imagine that addError were implemented such that it had access to enough context to null out _controller, so a second bang is necessary, but I don't think is supported in the language.

PR #658, opened 2020-11-01
Migrate async/iteration to non-null by default

cbracken:
Partial fix towards #606.

### DIFF
--- a/lib/src/async/iteration.dart
+++ b/lib/src/async/iteration.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:async';
 
 /// An asynchronous callback that returns a value.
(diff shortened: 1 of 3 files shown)


lib/src/async/iteration.dart
```diff
@@ -42,7 +40,7 @@ Future<S> _reduceAsync<S, T>(
 
 /// Schedules calls to [action] for each element in [iterable]. No more than
 /// [maxTasks] calls to [action] will be pending at once.
-Future<Null> forEachAsync<T>(Iterable<T> iterable, AsyncAction<Null, T> action,
```
cbracken, line 45: I wonder if we should just deprecate this in Quiver 2.0 and delete it here? Future.forEach](https://api.flutter.dev/flutter/dart-async/Future/forEach.html) is pretty much the same thing though it doesn't have the task limit, which I suppose could be useful for limiting things that involve system resources like file handles or sockets.

yjbanov: Do you mean 3.0? I'm OK with deprecating. I think `package:pool` is the de facto standard for this sort of thing.

yjbanov (approved):
lgtm

cbracken:
> Do you mean 3.0?

I mean deprecate on the main branch, release a new release, then delete here.

PR #659, opened 2020-11-01
Migrate async/stream_buffer to non-null by default

cbracken:
Partial fix towards #606.

### DIFF
--- a/lib/src/async/stream_buffer.dart
+++ b/lib/src/async/stream_buffer.dart
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:async';
 
 /// Underflow errors happen when the socket feeding a buffer is finished while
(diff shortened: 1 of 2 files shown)


lib/src/async/stream_buffer.dart
```diff
@@ -173,12 +170,10 @@ class StreamBuffer<T> implements StreamConsumer<List<T>> {
...
   @override
   Future close() {
-    Future ret;
-    if (_sub != null) {
-      ret = _sub.cancel();
-      _sub = null;
-    }
-    return ret ?? Future.value();
+    Future? ret;
+    ret = _sub?.cancel();
```
yjbanov, line 174: The previous 2 lines can be combined into one and `ret` can be made `final`.

yjbanov (approved):
lgtm

lib/src/async/stream_buffer.dart
cbracken: Thanks for spotting -- was done locally and forgot to re-push!

PR #662, opened 2020-11-03
Migrate collection/treeset to non-null by default

cbracken:
Partial fix towards #606.

### DIFF
--- a/lib/src/collection/treeset.dart
+++ b/lib/src/collection/treeset.dart
@@ -12,19 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @dart = 2.9
-
 import 'dart:collection';
 
 import 'package:meta/meta.dart' show visibleForTesting;
+
+int _defaultCompare(a, b) {
+  return a.compareTo(b);
+}
 
 /// A [Set] of items stored in a binary tree according to [comparator].
 /// Supports bidirectional iteration.
 abstract class TreeSet<V> extends IterableBase<V> implements Set<V> {
   /// Create a new [TreeSet] with an ordering defined by [comparator] or the
   /// default `(a, b) => a.compareTo(b)`.
-  factory TreeSet({Comparator<V> comparator}) {
-    comparator ??= (a, b) => (a as dynamic).compareTo(b);
+  factory TreeSet({Comparator<V> comparator = _defaultCompare}) {
     return AvlTreeSet(comparator: comparator);
   }
 
@@ -84,20 +85,25 @@ enum TreeSearch {
 /// A node in the [TreeSet].
 abstract class _TreeNode<V> {
   /// TreeNodes are always allocated as leafs.
-  _TreeNode({this.object});
+  _TreeNode({required this.object});
 
   _TreeNode<V> get left;
+  bool get hasLeft;
+
   _TreeNode<V> get right;
+  bool get hasRight;
 
   // TODO(codefu): Remove need for [parent]; this is just an implementation
   // note.
   _TreeNode<V> get parent;
+  bool get hasParent;
+
   V object;
 
   /// Return the minimum node for the subtree
   _TreeNode<V> get minimumNode {
     var node = this;
-    while (node.left != null) {
+    while (node.hasLeft) {
       node = node.left;
     }
     return node;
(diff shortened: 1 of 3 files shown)


cbracken:
I'll be honest, this is the migration I'm least confident in. There were a couple cases where TreeSet's implementation didn't behave according to the Set API contract. Iterator.current is now specified to return non-null for non-null generic set types and has been updated to throw in cases where it's called prior to moveNext or after moveNext has returned false.

lib/src/collection/treeset.dart
```diff
@@ -645,76 +645,83 @@ class AvlTreeSet<V> extends TreeSet<V> {
...
   @override
   V get first {
-    if (_root == null) return null;
-    AvlNode<V> min = _root.minimumNode;
-    return min != null ? min.object : null;
+    _TreeNode<V>? min = _root?.minimumNode;
+    if (min != null) {
+      return min.object;
+    }
+    throw StateError('No first element');
```
davidmorgan, line 652: What if `V` is nullable and the first element is `null`?

lib/src/collection/treeset.dart
```diff
@@ -645,76 +645,83 @@ class AvlTreeSet<V> extends TreeSet<V> {
...
   @override
   V get last {
-    if (_root == null) return null;
-    AvlNode<V> max = _root.maximumNode;
-    return max != null ? max.object : null;
+    _TreeNode<V>? max = _root?.minimumNode;
+    if (max != null) {
+      return max.object;
+    }
+    throw StateError('No last element');
```
davidmorgan, line 662: What if `V` is nullable and the last element is `null`?

lib/src/collection/treeset.dart
```diff
@@ -929,17 +938,22 @@ class _AvlTreeIterator<V> implements BidirectionalIterator<V> {
...
+  late int state;
+  _TreeNode<V>? _current;
 
   @override
-  V get current => (state != WALK || _current == null) ? null : _current.object;
+  V get current {
+    if (state == WALK && _current != null) {
+      return _current!.object;
+    }
+    throw StateError('No current element');
```
davidmorgan, line 955: It's possible to make this change in a less breaking way; e.g. v2/sdk/lib/internal/iterable.dart:

```
    // Before NNBD, this returned null when iteration was complete. In order to
    // avoid a hard breaking change, we return "null as E" in that case so that
    // if strong checking is not enabled or E is nullable, the existing
    // behavior is preserved.
    if (_remaining < 0) return null as E;
    return _iterator.current;
  }
```

davidmorgan (approved):
LG but a few suggestions. Some tricky bits to this one :)

jtmcdole:
Taking a quick look

lib/src/collection/treeset.dart
```diff
@@ -84,56 +85,61 @@ enum TreeSearch {
...
   _TreeNode<V> get parent;
+  bool get hasParent;
+
   V object;
 
   /// Return the minimum node for the subtree
   _TreeNode<V> get minimumNode {
-    var node = this;
-    while (node.left != null) {
+    _TreeNode<V> node = this;
```
jtmcdole, line 105: was the type on the left required? I thought it would get the type form "this"?

jtmcdole: same in a couple of places below. I'm OK with "yeah, NNBD requires the type so it can't be null".

lib/src/collection/treeset.dart
jtmcdole: I'm a little confused - the only null check here is on the Node; which must not be null. The object it carries (V) can be null and is returned at line 650.  

Or are you suggesting the signature should be `V?`

jtmcdole (approved):

lib/src/collection/treeset.dart
davidmorgan: My mistake, mixed up the node with the value. SG.

lib/src/collection/treeset.dart
davidmorgan: (ignore, parse failure on my part).

lib/src/collection/treeset.dart
cbracken: There are a lot of assumptions here that preclude supporting nullable values -- perhaps the answer is to set the type to `V extends Object` for now then make everything safe for null objects.

lib/src/collection/treeset.dart
cbracken: Done.

lib/src/collection/treeset.dart
cbracken: I can change it back to var. Added the type since during migration I had loosened it to `_TreeNode<V>?` while before I addded hasLeft/hasRight and a non-null return type on left/right. var is fine here.

cbracken: Done.

PR #668, opened 2020-11-05
Allow null args to lists/maps/setsEqual

cbracken:
This updates the parameters to listsEqual, mapsEqual and setsEqual in
the collection utils library to nullable types.

This avoids forcing callers to manually check for null, is a bit more in
line with Object.equals() which handles the null check in the runtime,
and avoids a breaking change to users who have not yet transitioned to
non-null by default.

Related: #606.

### DIFF
--- a/lib/src/collection/utils.dart
+++ b/lib/src/collection/utils.dart
@@ -17,8 +17,9 @@
 /// Returns `true` if [a] and [b] are both null, or they are the same length
 /// and every element of [a] is equal to the corresponding element at the same
 /// index in [b].
-bool listsEqual(List a, List b) {
+bool listsEqual(List? a, List? b) {
   if (a == b) return true;
+  if (a == null || b == null) return false;
   if (a.length != b.length) return false;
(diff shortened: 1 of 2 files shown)


lib/src/collection/utils.dart
```diff
@@ -49,8 +51,9 @@ bool mapsEqual(Map a, Map b) {
 ///
 /// Returns `true` if [a] and [b] are both null, or they are the same length and
 /// every element in [b] exists in [a].
-bool setsEqual(Set a, Set b) {
+bool setsEqual(Set? a, Set? b) {
   if (a == b) return true;
+  if (a == null || b == null) return false;
```
yjbanov, line 56: Do we need to test the null cases?

yjbanov (approved):
lgtm

lib/src/collection/utils.dart
cbracken: Yep good call -- done.

PR #669, opened 2020-11-05
Deprecate forEachAsync, reduceAsync

cbracken:
Along with doWhile, these methods can be directly replaced with
Future.forEach or Future.doWhile from the Dart core libraries. Added
migration code samples to all three functions.


yjbanov (approved):
lgtm

PR #671, opened 2020-11-06
Eliminate dependency on package:meta

cbracken:
This eliminates the dependency on package:meta, which was previously only used for `@required` and `@visibleForTesting`. 

`@required` was eliminated with the transition to non-null by default.

`@visibleForTesting` was eliminated by replacing `AvlTree.getNode(object)` with `debugGetNode(AvlTree, object)` and hiding it and `AvlNode` in the top-level re-export in the collection library.


yjbanov (approved):
lgtm

PR #672, opened 2020-11-06
Deprecate Optional

cbracken:
With the introduction of non-null by default in Dart SDK 2.12, existing
users should migrate to non-nullable types. This type will be removed in
Quiver 4.0.0.

See: #666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 #### 3.2.0-dev
 
+  * Deprecate `Optional`. With the introduction of non-null by default in Dart
+    SDK 2.12, existing users should migrate to non-nullable types. This type
+    will be removed in Quiver 4.0.0.
   * Expose `TreeIterator` for iterating `TreeSet` instead of using
     `BidirectionalIterator`.
   * Require Dart 2.17
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -24,7 +24,9 @@ import 'dart:collection';
 ///
 /// With the introduction of non-null by default in Dart SDK 2.12, developers
 /// should avoid adding more uses of this type. Existing users should migrate
-/// away from the `Optional` type to types marked nullable: `T?`.
+/// away from the `Optional` type to types marked nullable: `T?`. This type
+/// will be removed in Quiver 4.0.0.
+@Deprecated('Migrate to a non-nullable type. Will be removed in 4.0.0')
 class Optional<T> extends IterableBase<T> {
   /// Constructs an empty Optional.
   const Optional.absent() : _value = null;
--- a/test/core/optional_test.dart
+++ b/test/core/optional_test.dart
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// ignore_for_file: deprecated_member_use_from_same_package
 library quiver.core.optional_test;
 
 import 'package:quiver/src/core/optional.dart';


cbracken:
From offline discussion with @yjbanov, we may want to delay landing this until packages like `built_*` have been migrated off Optional to avoid filling downstream users' analysis with deprecation warnings.

What I may do is update this to simply update the docs to suggest people avoid adding new dependencies on it, then add the attribute once the major downstream packages have been migrated off.

davidmorgan:
FWIW `built_` never used/recommended `Optional`, in anticipation of null safety :)

cbracken:
Oh - fantastic news!

yjbanov (approved):
lgtm

cbracken:
I've rebased to tip of tree. Once NNBD support is enabled across the board internally, I'll:
1. Do a small handful of manual migrations to identify any unforseen issues
2. Write up a short migration HOWTO doc
3. If it looks feasible, use the large-scale change process to notify everyone, perform the migration, then regret having signed up for it when I get to the final, painful 5% that hits every edge-case I missed in step 1. :)

modulovalue:
> then regret having signed up for it when I get to the final, painful 5% that hits every edge-case I missed in step 1. :)

Null safe types are not a replacement for an `Optional` type.
You can't nest nullables, but you can nest Optionals.
i.e. `int??` is not allowed in Dart, but `Optional<Optional<int>>` is.

Optional has a reason to exist. 
Think of e.g. an API call that returns a nullable bool. To model its absence an Optional type would be necessary.

_I just randomly stumbled across this PR, don't mind me, but you might want to check for that edge-case or decide to keep Optionals :)_

cbracken:
> Think of e.g. an API call that returns a nullable bool. To model its absence an Optional type would be necessary.

Can you explain a bit further? `bool`, `int`, etc. in Dart (prior to NNBD) are all nullable types. To model a nullable `bool` post-NNBD, you'd type it as `bool?` instead.

hoc081098:
```dart
Map<String, Optional<User>> cache; // user by id

// there 3 cases:
cache['some_id'] == null; // -> does not contains in cache
cache['some_id]'.isPresent == true; // -> cache contains a User
cache['some_id'].isAbsent == true; // -> User not exists
``` 

_Sent from my Redmi  7A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

cbracken:
@hoc081098 you can store a null value in maps today.
```dart
Map<String, User?> cache;
cache.containsKey('some_id') == false; // -> not contained in cache
cache.containsKey('some_id') == true; // -> cache contains a user
```

In general, it is unsafe to assume that `cache[key] == null` is equivalent to `cache.containsKey(key)`, since maps allow null values.

modulovalue:
@cbracken 
The following is valid Swift code:
```swift
let i: Int????? = nil // the variable i has the quiver-equivalent type Optional<Optional<Optional<...int>>>
```

But that above  is not expressible in dart without a wrapper
```dart
int???? i = null; // compile time error
```

To simulate the way that nulls work in swift, users would have to declare an Optional wrapper themselves.

In other words, the Optional type allows users to define a nullable nullable value.

An example where that could be useful can be seen here https://github.com/dart-lang/sdk/blob/8432b379e7fd2fae1e872ca20f9275366d49ffe0/pkg/vm_service/lib/src/dart_io_extensions.dart#L184-L185
The vm_service api has collapsed their union types into single types where values that can be null can be absent. (Optional<int?> would be the best way to model that since int?? is not valid dart)

hoc081098:
I think nested optional type is a good point to keep `Optional` 😃

modulovalue:
@cbracken You can ignore my comments, I was wrong. I assumed Optional behaved more like a Maybe/Option Monad but it doesn't at all.

cbracken:
I'm going to close this. As much as it makes sense to

cbracken:
Based on comments in b/258898406 internally, I'm going to rebase and land this.

PR #673, opened 2020-12-21
Make a StreamRouter's route remove itself when all listener cancel

matehat:
In my [dart library][1] where we use StreamRouter, some weird bug was exhibited where a channel being previously listened for, then cancelled, then re-listened would get messages on the second listening.

That was because, basically, if you don't reuse the initial stream you got the first time you called `StreamRouter().route(...)`, subsequent equivalent routing would never receive events from the stream router. That's expected, unless you cancel all listeners from the initial stream. 

With this PR, once all listeners to the stream have cancelled, the corresponding route is closed and removed from the stream router. This also has the benefit that a stream router now automatically scales down with the amount of stream listened to.


matehat:
@googlebot I signed it!

PR #678, opened 2021-02-22
Fix error when AVLTree root element is removed

cbracken:
When removing the root element of an AVLTree, the parent is null. The
getter applies the ! operator to node._parent. Instead we reference the
member directly.


jtmcdole (approved):
LGTM

PR #684, opened 2021-06-14
Migrate from dartanalyzer to dart analyze

cbracken:
In Dart SDK 2.10, the `dart analyze` command was added as the
recommended way to run Dart source code analysis.


yjbanov (approved):
lgtm

PR #688, opened 2021-10-14
migrated `Optional<T>` to `T?`

### LINKED ISSUE #679: migrate Optional<T> -> "T?"
Hi thanks for the lib! Since now null safety is stable, we want to migrate all `Optional<T>` to `T?`. Can this be done with some tools? Thanks

adityatheoctocatdev:
### fixes #679


cbracken:
It's worth noting that #679 isn't a feature request for an API change in quiver, but rather a request for documentation on how to migrate off of `Optional` to `T?` instead. My comments there are intended to help warn developers about common traps that make such a migration difficult.

PR #689, opened 2021-10-14
Version 3.0.1+1

cbracken:
This version adds a minor documentation change.

With non-null by default implemented as of Dart 2.12, developers should
avoid adding more uses of the Optional type in there code, and ideally
start migrating existing uses to nullable types.

Not marking this as deprecated yet, since not all users have migrated
their codebases to NNBD, and migration off this type has some tricky
edge-cases.

Related: #679
Related: #672 (deprecate optional)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+#### 3.0.1+1 - 2021-10-14
+
+  * Add documentation to `Optional` suggesting that adding new uses be avoided
+    and existing uses should be migrated to nullable types in codebases where
+    non-null by default has been enabled.
+
 #### 3.0.1 - 2021-04-06
 
   * Fix: Eliminate null check error on removal of root node of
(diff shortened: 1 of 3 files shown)


yjbanov (approved):
lgtm

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+#### 3.0.1+1 - 2021-10-14
```
natebosch, line 1: FYI `3.0.2` would have been more idiomatic.

https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#bumping-versions

cbracken: Oh weird. TIL! 

Out of curiosity, when did this rule change? We'd been doing this for years, as had other packages like vm_service, dart_style, async, etc.

natebosch: There was a bunch of confusion years ago about the intended patterns. Some packages only used `+1` version changes for `0.x` releases, while other packages kept using them after the `1.x` release for doc changes.

cbracken: SGTM - glad we've got a clearly documented policy now!

Issue #693, opened 2022-05-03
Cache.get's optional ifAbsent parameter should be nullable


cbracken:
During non-null by default migration, `Cache.get`'s `ifAbsent` parameter was incorrectly marked non-nullable during non-null by default migration. The parameter is optional and the doc comment clearly states that `ifAbsent` is not required.

PR #696, opened 2022-05-03
Make Cache.get ifAbsent param nullable

cbracken:
During non-null by default migration, Cache.get's ifAbsent parameter was
incorrectly marked non-nullable. The parameter is optional and the doc
comment clearly states that ifAbsent is not required.

Reported internally at Google by Yutong Jin.

Issue: https://github.com/google/quiver-dart/issues/693


stereotype441:
lgtm

PR #700, opened 2022-06-23
Stop exposing `BidirectionalIterator`.

lrhn:
I'm planning to deprecate `BidirectionalIterator`, hopefully to remove it in Dart 3.0.

The `TreeSet` exposes a bidirectional iterator. 
By changing it to making the `_AvlTreeIterator` public as `TreeIterator`, the API no longer depends on `BidirectionalIterator`.

The `_AvlTreeIterator` exposed some of its state as public getters. I have made some of them private, and kept others as public. 
If they weren't intended to be public, we can make them all private.

Code which uses `.iterator` with inference will now infer `TreeIterator` instead, with (at least) the same API.
Code which uses `BidirectionalIterator` will still work, because the `TreeIterator` still implements the interface, but will get their own deprecation warnings so they can change it to `TreeIterator`.

In the long run, I'd recommend just implementing `Iterator`, and let `TreeIterator` define the `movePrevious` itself.

### DIFF
--- a/lib/src/collection/treeset.dart
+++ b/lib/src/collection/treeset.dart
@@ -40,22 +40,22 @@ abstract class TreeSet<V> extends IterableBase<V> implements Set<V> {
   @override
   bool get isNotEmpty => length != 0;
 
-  /// Returns an [BidirectionalIterator] that iterates over this tree.
+  /// Returns an [TreeIterator] that iterates over this tree.
   @override
-  BidirectionalIterator<V> get iterator;
+  TreeIterator<V> get iterator;
 
-  /// Returns an [BidirectionalIterator] that iterates over this tree, in
+  /// Returns an [TreeIterator] that iterates over this tree, in
   /// reverse.
-  BidirectionalIterator<V> get reverseIterator;
+  TreeIterator<V> get reverseIterator;
 
-  /// Returns an [BidirectionalIterator] that starts at [anchor].  By default,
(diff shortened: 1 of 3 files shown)


lib/src/collection/treeset.dart
```diff
@@ -40,22 +40,22 @@ abstract class TreeSet<V> extends IterableBase<V> implements Set<V> {
   @override
   bool get isNotEmpty => length != 0;
 
-  /// Returns an [BidirectionalIterator] that iterates over this tree.
+  /// Returns an [TreeIterator] that iterates over this tree.
```
cbracken, line 43: nit (here and below): "a" TreeIterator now.

lib/src/collection/treeset.dart
```diff
@@ -440,7 +440,7 @@ class AvlTreeSet<V> extends TreeSet<V> {
     ++_modCount;
     --_length;
 
-    // note: if you read wikipedia, it states remove the node if its a leaf,
+    // note: if you read wikipedia, it _states remove the node if its a leaf,
```
cbracken, line 443: Revert back to `states`.

cbracken (approved):
Thanks for sending the fixup, Lasse! Looks great; we'll follow up with the removal.

Just a few find-replace related nits :)

lib/src/collection/treeset.dart
lrhn: Done.

PR #701, opened 2022-07-25
Remove deprecated experimental invariant_booleans lint rule

srawlins:
CC @kevmoo

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -71,7 +71,6 @@ linter:
     - flutter_style_todos
     - hash_and_equals
     - implementation_imports
-    # - invariant_booleans # too many false positives: https://github.com/dart-lang/linter/issues/811
     - iterable_contains_unrelated_type
     # - join_return_with_assignment # not yet tested
     - library_names


analysis_options.yaml
```diff
@@ -71,7 +71,6 @@ linter:
     - flutter_style_todos
     - hash_and_equals
     - implementation_imports
-    # - invariant_booleans # too many false positives: https://github.com/dart-lang/linter/issues/811
```
kevmoo, line 74: Ha! False positive. Although you uncovered the bidirectional iterator silly @srawlins !

kevmoo (approved):

cbracken (approved):

analysis_options.yaml
cbracken: BidirectionalIterator cleanup landed in https://github.com/google/quiver-dart/pull/700.

Now that it's deprecated, we can probably stop implementing it in `TreeIterator`. I'll send a patch and bump the major version since it'll be a breaking change.

PR #703, opened 2022-09-07
Update all copyright header dates for consistency

cbracken:
As done with the Dart SDK and Flutter, this updates Quiver to set all copyright headers to the earliest value for consistency. Any future files should also set their dates to 2013.

Issue: https://github.com/google/quiver-dart/issues/702


yjbanov (approved):
lgtm

PR #706, opened 2022-11-10
docs: rename IsochronousStream to Metronome

korzonkiee:
When reading the docs for the `Metronome` class I noticed that there is a reference to a `IsochronousStream` class that doesn't exist in the quiver library. I think should be renamed to `Metronome`. Also, the parameter is now `anchor` rather than `anchorMs` 😁.


cbracken:
Amazing. That comment has been there for years and no one's spotted it till now! Thanks for fixing!

jtmcdole:
totally wasn't my fault... no one liked `IsochronousStream` :)

PR #711, opened 2022-11-15
Correctly handle nullable type arguments in a number of APIs

kevmoo:


lib/src/collection/bimap.dart
```diff
@@ -151,7 +151,7 @@ class HashBiMap<K, V> implements BiMap<K, V> {
   @override
   void updateAll(V Function(K key, V value) update) {
     for (final key in keys) {
-      _add(key, update(key, _map[key]!), true);
+      _add(key, update(key, _map[key] as V), true);
```
kevmoo, line 154: handles the nullable `V` case correctly!

This could also iterate over `entries` to avoid the `as` check...

lib/src/collection/lru_map.dart
```diff
@@ -290,7 +290,7 @@ class LinkedLruHashMap<K, V> implements LruMap<K, V> {
   V update(K key, V Function(V value) update, {V Function()? ifAbsent}) {
     V newValue;
     if (containsKey(key)) {
-      newValue = update(this[key]!);
+      newValue = update(this[key] as V);
```
kevmoo, line 293: No way around this, if I understand things correctly

lib/src/core/optional.dart
```diff
@@ -23,7 +23,7 @@ import 'dart:collection';
 /// With the introduction of non-null by default in Dart SDK 2.12, developers
 /// should avoid adding more uses of this type. Existing users should migrate
 /// away from the `Optional` type to types marked nullable: `T?`.
-class Optional<T> extends IterableBase<T> {
+class Optional<T extends Object> extends IterableBase<T> {
```
yjbanov, line 26: This is probably breaking, but if you can pull this through, hats off to you!

yjbanov (approved):

PR #714, opened 2022-12-08
Stop implementing deprecated `BidirectionalIterator`.

lrhn:
The `TreeIterator` implemented `BidirectionalIterator`, which is deprecated and will go away in Dart 3.0.

This changes it to just implement `Iterarator`, and have its own `movePrevious`.

This could be a breaking change, but it does not appear that anyone ever casts the `TreeIterator` to `BidirectionalIterator` (because they might as well just keep the `TreeIterator` type, and there is no helper function anywhere in the world which expects just a `BidirectionalIterator` that you can pass the iterator to).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
   * Deprecate `Optional`. With the introduction of non-null by default in Dart
     SDK 2.12, existing users should migrate to non-nullable types. This type
     will be removed in Quiver 4.0.0.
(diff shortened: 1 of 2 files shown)


mit-mit:
cc @kevmoo

kevmoo (approved):

mit-mit:
@kevmoo can you merge too? we don't have the needed permissions

PR #727, opened 2023-04-19
Add factory constructor Multimap.fromMultimap

### LINKED ISSUE #705: Implement a factory constructor on Multimap that copies another Multimap
Map has .from() and .of() factory constructors, why doesn't Multimap?

patrikbadanjak:
A little convenience method for easier copying. Closes #705.

### DIFF
--- a/lib/src/collection/multimap.dart
+++ b/lib/src/collection/multimap.dart
@@ -30,6 +30,10 @@ abstract class Multimap<K, V> {
       {K Function(dynamic)? key,
       V Function(dynamic)? value}) = ListMultimap<K, V>.fromIterable;
 
+  /// Constructs a new multimap with the same values as [other]
+  factory Multimap.fromMultimap(Multimap<K, V> other) =
+      ListMultimap<K, V>.fromMultimap;
+
   /// Returns whether this multimap contains the given [value].
   bool containsValue(Object? value);
 
@@ -117,6 +121,11 @@ abstract class _BaseMultimap<K, V, C extends Iterable<V>>
     for (final element in iterable) {
       add(key(element), value(element));
     }
+  }
+
+  /// Constructs a new multimap with the same values as [other]
+  _BaseMultimap.fromMultimap(Multimap<K, V> other) {
+    addAll(other);
   }
 
   static T _id<T>(x) => x;
@@ -239,6 +248,9 @@ class ListMultimap<K, V> extends _BaseMultimap<K, V, List<V>> {
   /// default to the identity function.
   ListMultimap.fromIterable(super.iterable, {super.key, super.value})
       : super.fromIterable();
+
+  /// Constructs a new multimap with the same values as [other]
+  ListMultimap.fromMultimap(super.other) : super.fromMultimap();
 
   @override
   List<V> _create() => [];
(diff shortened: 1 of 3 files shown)


lib/src/collection/multimap.dart
```diff
@@ -30,6 +30,10 @@ abstract class Multimap<K, V> {
       {K Function(dynamic)? key,
       V Function(dynamic)? value}) = ListMultimap<K, V>.fromIterable;
 
+  /// Constructs a new multimap with the same values as [other]
```
cbracken, line 33: Please add a period at the end of the doc comment.

Ref: https://dart.dev/guides/language/effective-dart/documentation

lib/src/collection/multimap.dart
```diff
@@ -119,6 +123,11 @@ abstract class _BaseMultimap<K, V, C extends Iterable<V>>
     }
   }
 
+  /// Constructs a new multimap with the same values as [other]
```
cbracken, line 126: Similar comment.

lib/src/collection/multimap.dart
```diff
@@ -240,6 +249,9 @@ class ListMultimap<K, V> extends _BaseMultimap<K, V, List<V>> {
   ListMultimap.fromIterable(super.iterable, {super.key, super.value})
       : super.fromIterable();
 
+  /// Constructs a new multimap with the same values as [other]
```
cbracken, line 252: And again here.

test/collection/multimap_test.dart
```diff
@@ -64,6 +64,41 @@ void main() {
...
+
+    test('should have same values', () {
+      var map1ContainsAllMap2Values =
+          mmap1.values.every((value) => mmap2.containsValue(value));
+      var map2ContainsAllMap1Values =
+          mmap2.values.every((value) => mmap1.containsValue(value));
+
+      expect(true, map1ContainsAllMap2Values && map2ContainsAllMap1Values);
+    });
+  });
```
cbracken, line 100: We should add a test to ensure no one inadvertently breaks the behaviour by, say, re-using the Lists/Sets we create internally in mmap1 -- either by modifying the factory constructor or by modifying addAll. I haven't looked to see that we already have this test for addAll (hopefully), but either way, it's worth being paranoid here.

Something like:
```dart
mmap1 = Multimap()
  ..add('k1', 'v1')
  ..add('k1', 'v2')
  ..add('k2', 'v3');
mmap2 = Multimap.fromMultimap(mmap1);
mmap1['k1'].remove('v1');
expect(mmap2['k1'].contains('v1'));
```

cbracken (commented):
Thanks for your patch! Just a few comments.

cbracken (approved):

patrikbadanjak:
I right clicked on the files in Android Studio and clicked on 'reformat code with dart format'. Then I ran it from the terminal and it produced a different output...

cbracken:
Interesting -- sounds like maybe a bug in the intellij/android studio integration. If it's reproducible, you could file a bug against https://github.com/dart-lang/sdk.

cbracken:
Thanks again for the patch!

patrikbadanjak:
The terminal uses the homebrew version unlike Android Studio which uses the one from the Flutter package. They have the same version code.

cbracken:
Ah - I'm using the commandline version from the SDK included in Flutter and it matches the corrected format but I'll bet that's because my flutter install is always near tip-of-tree on master branch since working on flutter is my day job. We did some significant updates for Dart 3 between the last stable release and now so perhaps that accounts for the difference.

Thanks for the additional info!

PR #731, opened 2023-06-06
Deprecate Optional

grouma:
We do not intend on fully removing this type given the large amount of internal code that depends on it. Also this abstraction is useful in rare instances for the null-safety migration. To discourage adoption however, we will deprecate it and update the documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 3.2.2-dev
 
 - Fix a number of doc comment issues.
+- Re-deprecate `Optional` but provide more nuance to the documentation.
 
 ## 3.2.1 - 2022-12-20
 
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -24,10 +24,11 @@ import 'dart:collection';
 ///
 /// With the introduction of non-null by default in Dart SDK 2.12, developers
 /// should avoid adding more uses of this type. Existing users should migrate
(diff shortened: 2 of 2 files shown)


kevmoo (approved):

cbracken (approved):

davidmorgan:
Thanks!

The new version looks good but while we're here we might as well quickly discuss aiming for great ;) ...

I would perhaps drop the first paragraph (before the changes), slightly soften "should migrate" and give examples of rare valid new use cases.

How about:

```
/// A value that might be absent.
///
/// This class was written before Dart SDK 2.12 added first class support for 
/// non-nullable types. Where possible, it is strongly preferred to use those   
/// language features and to not use this class at all.
///
/// Usually, `Optional<T>` should be migrated to `T?`.
///
/// However, code that already uses `Optional` can continue to use it if the
/// cost of migration is considered too high. The class will not be removed.
///
/// Valid new use cases:
///
/// - If an API only accepts non-null values, you can use `Optional` as a
///   general way to pass it values that mean "absent";
/// - for a general way to represent "nested levels" of absent, such as
///   `Optional<Optional<T>>`.
///
/// Both of these should be very rare.
@Deprecated('Generally, migrate to a nullable type.')
```

If this looks better or worth discussion I'll turn it into a PR; if not, I'm happy with the version that just went in.

Please let me know.

Thanks everyone :)

PR #735, opened 2023-07-10
feat: add letter judgment for String

drown0315:
Add letter judgment for String

### DIFF
--- a/lib/strings.dart
+++ b/lib/strings.dart
@@ -159,9 +159,9 @@ bool equalsIgnoreCase(String? a, String? b) =>
 int compareIgnoreCase(String a, String b) =>
     a.toLowerCase().compareTo(b.toLowerCase());
 
-/// if [s] is consists of only letters.
-bool isLetters(String s){
-   return RegExp(r'^[A-Za-z]+$').hasMatch(s);
+/// Return true if [s] is consists of only letters.
+bool isLetters(String s) {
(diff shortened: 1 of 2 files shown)


cbracken (changes requested):
I'm going to close this PR since this isn't something that we're likely to add to quiver.

`isDigit` is pretty straightforward since there are only 10 of them to test against and developers mostly agree that they're the ones in UTF-8 range 0x30-0x39. There are some very common use-cases for such a function, such as verifying that a user has input a numeric value, for example. While some languages do support other digits like '４' (full-width 4 in Japanese) or 四, in numeric use-cases like spreadsheets, almost all software uses 0-9 for numerics.

`isLetters`, on the other hand, is pretty ambiguous and locale-dependent, and there is a wide variety of slightly differing use-cases for this category function. 

- Should it include diacritics, which are necessary for many languages (including English) for words like café and names like Emily Brontë?
- Should it include alphabetic letters that don't appear in English like the [dotless I](https://en.wikipedia.org/wiki/Dotless_I) in Turkish?
- Should it exclude the letters J, K, W, X, Y if the user's locale is Italian, whose alphabet doesn't include those letters?
- Should non-latin characters like μ or あ or Ю be included?

I can imagine legitimate use-cases for wanting to check if a word is composed of only a-z/A-Z, but I imagine they're a very small fraction of all cases where a developer wants to check a string, so it's something I think we should leave to individual app developers to handle.

PR #736, opened 2023-07-10
Change definition to `class Optional<T extends Object>`

oprypin:
This prevents `Optional` from being declared with a nullable type. It is never useful to do so, because trying to create an instance with a `null` causes a run-time error anyway. So just confusing.

Internally we already applied this change and were quite easily able to clean up any existing occurrences.

### DIFF
--- a/lib/src/core/optional.dart
+++ b/lib/src/core/optional.dart
@@ -36,7 +36,7 @@ import 'dart:collection';
 ///
 /// Both of these should be very rare.
 @Deprecated('Generally, migrate to a nullable type.')
-class Optional<T> extends IterableBase<T> {
+class Optional<T extends Object> extends IterableBase<T> {
   /// Constructs an empty Optional.
   const Optional.absent() : _value = null;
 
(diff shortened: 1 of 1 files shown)


cbracken:
See: https://github.com/google/quiver-dart/pull/667

oprypin:
> See: https://github.com/google/quiver-dart/pull/667

> A survey of the Google codebase suggests this would be a fairly extensive breaking change.

It *is* a breaking change but as I said I already applied this change there. So if Google codebase was a big part of the reason, it no longer applies.

cbracken:
Apologies -- submitted the comment accidentally while I was still writing it. It's updated now.

oprypin:
> It may be useful to conduct a quick use of open source projects to see what impact we can expect to users.

Hm I haven't done such a survey before. Maybe if someone already has a setup for something like this, they could chip in.

cbracken:
@yjbanov I think you've got experience doing this previously using BigQuery? If not, it's something the Flutter framework team has a lot of experience with. Will see if I can dig up a doc.

Issue #749, opened 2024-08-28
Can't access most of API documentation (404)


curiouslaw:
When I try to access the API Documentation, most of them shows 404 (e.g.,: https://pub.dev/documentation/quiver/latest/quiver.check/quiver.check-library.html). 

Out of the 15 libraries, only 3 are not broken (cache, pattern, and string).

cbracken:
@kevmoo who's the right person for pub doc links being broken?

kevmoo:
maybe Sam Rawlins? Perhaps there is a dartdoc bug?

srawlins:
Yeah that could definitely be a dartdoc bug...

From https://pub.dev/documentation/quiver/latest/, clicking "quiver.async" takes me to https://pub.dev/documentation/quiver/latest/quiver.async/quiver.async-library.html and a 404.

And sure enough, trying to document quiver locally, I get a stack overflow 😱 . Good catch.

cbracken:
@srawlins mentions this may be due to the dartdoc for `Multimap.fromIterable`, which is as follows:
```dart
  /// Constructs a new list-backed multimap. For each element e of [iterable],
  /// adds an association from [key](e) to [value](e). [key] and [value] each
  /// default to the identity function.
  factory Multimap.fromIterable(Iterable iterable,
      {K Function(dynamic)? key,
      V Function(dynamic)? value}) = ListMultimap<K, V>.fromIterable;
```

The intent here is that the key and `key` and `value` params. These are tagged as references to the params in the docs as `[key]` and `[value]`, but the docs are explaining that for each element `e` in `iterable` we're inserting a mapping from `key(e)` to `value(e)`. Unfortunately, referring to the function params by bracketing them makes them look like markdown links.

cbracken:
3.2.2 is published. The [docs](https://pub.dev/documentation/quiver/3.2.2/) appear to be there now.

Thanks for reporting @curiouslaw.

PR #750, opened 2024-08-28
Fix some doc comments

srawlins:
Should ultimately lead to the fix of https://github.com/google/quiver-dart/issues/749, when quiver 3.2.2 is released.

### DIFF
--- a/lib/src/collection/multimap.dart
+++ b/lib/src/collection/multimap.dart
@@ -23,9 +23,10 @@ abstract class Multimap<K, V> {
   /// Constructs a new list-backed multimap.
   factory Multimap() => ListMultimap<K, V>();
 
-  /// Constructs a new list-backed multimap. For each element e of [iterable],
-  /// adds an association from [key](e) to [value](e). [key] and [value] each
-  /// default to the identity function.
+  /// Constructs a new list-backed multimap.
+  ///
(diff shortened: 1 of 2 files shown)


lib/src/collection/multimap.dart
```diff
@@ -23,9 +23,10 @@ abstract class Multimap<K, V> {
   /// Constructs a new list-backed multimap.
   factory Multimap() => ListMultimap<K, V>();
 
-  /// Constructs a new list-backed multimap. For each element e of [iterable],
-  /// adds an association from [key](e) to [value](e). [key] and [value] each
-  /// default to the identity function.
+  /// Constructs a new list-backed multimap.
+  ///
+  /// For each element `e` of [iterable], adds an association from [key] to
+  /// [value]. [key] and [value] each default to the identity function.
```
cbracken, line 29: This is not what the constructor does. We're mapping from `key(e)` to `value(e)`.

cbracken (approved):
Added a note.

cbracken (changes requested):
Sorry that should have been request changes.

lib/src/collection/multimap.dart
srawlins: Updated, thanks!

cbracken (approved):

Thanks!

PR #757, opened 2025-02-04
Deprecate hash functions

kevmoo:
Folks should use Object.hash bits instead
Replaced impl w/ Object.hash bits


cbracken (approved):

